PR #29, opened 2021-07-07
feat(qna): extract module qna on studio

allardy:
This PR brings the frontend and backend part of the QNA module used to manage files. There is also a small router and service that will be added on the main repository so the middleware can be active.

Core:
- Add route to get all questions, a specific question, or intent actions (for NDU)

Studio:
- Moved all UI to the studio
- Moved "actions" from the core to studio (eg: when flow is renamed changed, or a topic is changed).
- Moved all related backend to support the UI
- Also moved NLU management routes from the core to studio

The end goal is that we can run the studio to create and manage all our bots, without having to load modules


EFF:
Link to QnA in debugger will probably be broken.
Same as for NLU

PR #30, opened 2021-07-12
fix(debugger): use message ids

samuelmasse:
This fixes the debugger to work with external messaging. Since messaging doesn't know about events, it can't list the messages along with the event ids. So we get the event corresponding to the message id that we have. 

Related to https://github.com/botpress/botpress/pull/5183


packages/studio-ui/src/web/components/Layout/BottomPanel/Debugger/index.tsx
```diff
@@ -132,15 +132,19 @@ export class Debugger extends React.Component<Props, State> {
...
     const eventsCache = this.state.eventsCache
 
-    const existing = eventsCache.find(x => x.id === eventId)
+    const existing = eventsCache.find(x => x.messageId === messageId)
     if (existing) {
       return existing
     }
 
-    const { data: event } = await axios.get(`${window.BOT_API_PATH}/mod/extensions/events/${eventId}`)
+    const { data: event } = await axios.get(`${window.BOT_API_PATH}/messaging/message-to-event/${messageId}`)
```
EFF, line 147: Does this route really exist ?

samuelmasse: Yes implemented here https://github.com/botpress/botpress/pull/5183

samuelmasse:
@allardy This is required to merge #5183

EFF (approved):

PR #33, opened 2021-07-19
fix(core): add missing number util

laurentlp:
This PR fixes the google-speech module available in the main repository that is requiring this missing util.


laurentlp:
Alternative PR: https://github.com/botpress/botpress/pull/5199

PR #35, opened 2021-07-22
fix(studio): don't open node props when moving it

sadda11asm:
There was an annoying issue - when a user moves a node, it opens up a node props on the right. I added a check - if the note has been moving or not, and so I disable the props in the first case.


laurentlp (approved):
Tested and LGTM!

Can you please use a clone of the repo instead of a fork, please? It really helps with the code review!

PR #37, opened 2021-07-28
fix(studio): overlapping between search bar & node props side panel

sadda11asm:
Added another CSS style and choose that one whenever the side panel is open


allardy:
Can you include a screenshot of before/after so the change is easier to see? Thanks!

EFF:
The PR description isn't clear. Next time try to add more details and as @allardy said when it touches UI try to add before/after screencaps

sadda11asm:
oops sorry guys, missed your replies, will do your requests asap

sadda11asm:
**How it was**:
1) the search tab before nodePropsTab is opened
image
2) the search tab after nodePropsTab is opened (it is behind it)
image
**How it now**:
2) the search tab after nodePropsTab is opened 
image
When nodePropsTab is bigger:
image
@allardy @laurentlp @EFF here are the screenshots :)

allardy:
I'm sure there's a way to implement this using CSS instead of adding another library and passing around the width of an opened panel. Can you look into that ?

sadda11asm:
> I'm sure there's a way to implement this using CSS instead of adding another library and passing around the width of an opened panel. Can you look into that ?

Okay I will look into that

sadda11asm:
@allardy I made changes per your request! Could you please review it again?

packages/studio-ui/src/web/views/FlowBuilder/index.tsx
```diff
@@ -182,23 +182,27 @@ const FlowBuilder = (props: Props) => {
...
+      <div className={style.container}>
+        <div className={style.diagram}>
+          <Diagram
+            readOnly={readOnly}
+            showSearch={showSearch}
+            hideSearch={() => setShowSearch(false)}
+            handleFilterChanged={handleFilterChanged}
+            highlightFilter={highlightFilter}
+            mutexInfo={mutex}
+            showFlowNodeProps={props.showFlowNodeProps}
```
EFF, line 194: This property is new ? how does that affect the ui ?

packages/ui-shared/src/MainLayout/Toolbar/index.tsx
```diff
@@ -17,7 +17,7 @@ const Toolbar: FC<ToolbarProps> = props => {
         <Tabs currentTab={props.currentTab} tabChange={props.tabChange} shouldFloat tabs={props.tabs} />
       )}
       {!!props.buttons?.length && (
-        <NavbarGroup className={cx(style.buttons, 'toolbar-buttons')} align={Alignment.RIGHT}>
+        <NavbarGroup style={{ marginTop: '5px' }} align={Alignment.RIGHT}>
```
laurentlp, line 20: Move this into a CSS class.

packages/studio-ui/src/web/views/FlowBuilder/index.tsx
laurentlp: Is this props required?

laurentlp (commented):
Small comments other than that tested and LGTM!

packages/studio-ui/src/web/views/FlowBuilder/index.tsx
sadda11asm: oh, I think I forgot to delete it after some debugging

laurentlp (approved):
LGTM!!

PR #40, opened 2021-08-03
fix(studio): Ui changes

ptrckbp:
Just going through list of ui changes for QoL improvements. @allardy Please check if changing to default value of canOutsideClickClose is going to break something.

### DIFF
--- a/packages/ui-shared/src/Dialog/index.tsx
+++ b/packages/ui-shared/src/Dialog/index.tsx
@@ -6,7 +6,7 @@ import style from './style.scss'
 import { DialogProps } from './typings'
 
 export const Wrapper: FC<DialogProps> = props => {
-  let width: any;
+  let width: any
   if (props.size === 'sm') {
     width = 360
   } else if (props.size === 'md') {
--- a/packages/ui-shared/src/Dialog/index.tsx
+++ b/packages/ui-shared/src/Dialog/index.tsx
@@ -6,7 +6,7 @@ import style from './style.scss'
 import { DialogProps } from './typings'
 
 export const Wrapper: FC<DialogProps> = props => {
-  let width: any = undefined
+  let width: any;
   if (props.size === 'sm') {
     width = 360
   } else if (props.size === 'md') {
--- a/packages/ui-shared/src/Dialog/index.tsx
+++ b/packages/ui-shared/src/Dialog/index.tsx
@@ -24,7 +24,7 @@ export const Wrapper: FC<DialogProps> = props => {
     <Dialog
       className={style.dialog}
       transitionDuration={0}
-      canOutsideClickClose
+      canOutsideClickClose={false}
       canEscapeKeyClose
       enforceFocus={false}
       style={{ width, height: props.height }}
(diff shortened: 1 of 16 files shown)


ptrckbp:
Before / afters :

Modal buttons no longer stuck together, there is more spacing for modals, and overflow menus are shown above. All modals are now of a bigger size. 
modal-before
modal-after

Increased action parameter ratio to half half, to make it easier to see suggestions for beginners. 
action-parameters-before
action-parameters-resize-after

Actions in inspector where twitching due to a border-left on hover. Removed the border and increased the background color change contrast.
twitching-before
twitching-after

Tooltips no longer hiding text / offscreen: 
tooltip-before
tooltip-after

I just removed the help button that linked nowhere in the flow page.

allardy:
Wouldn't it make more sense to use icons instead? Text feels weird, at least if we use icons they will be aligned and be much easier to use

image

allardy:
Also, I think we should pick an "enlarged" size for the inspector, and stick to that size. Having the window larger or smaller depending on the size of the content results in a poor UX since you can't just move your mouse where you expect the tab to be once you click on a choice


image
image

allardy:
I'm not sure about the removal of the size of the modal... This one feels way too big for the content:
image

ptrckbp:
I removed as many modifications as possible, and made the "rename flow" dialog the same width as previously, but for unspecified sizes, the default is still Increased.  

I agree with your point about the inspector and would be happy to figure this out, but is my PR is causing that? I checked out master and saw the non-fixed width issue as well. 

Finally, I don't understand your comment about icons. :( Is it a blocking issue for this PR?

Let me know if you want any more changes!

EFF:
We have no ux designer so let's not debate icons. If @allardy you don't have any other comments on the actual implementation . this could be good to merge

packages/ui-shared/src/Dialog/index.tsx
```diff
@@ -24,13 +24,14 @@ export const Wrapper: FC<DialogProps> = props => {
     <Dialog
       className={style.dialog}
       transitionDuration={0}
-      canOutsideClickClose={false}
+      canOutsideClickClose
```
allardy, line 27: This actually is a regression. We purposely disabled clicking outside dialogs to close them since you may click accidentaly outside while filling a form (or writing a cntent element, for example) which made you lose your current work. 

So escape to close is good, clicking outside should not close it

allardy (commented):
There's a lot of dialogs everywhere and the change to change the default width from 500 to 750 make some of those dialogs feel weird (being way too big for their contents), ex:
image

Shouldn't we keep the same dialog size, but expand those which should be bigger, like the content or flow items?

packages/ui-shared/src/Dialog/index.tsx
ptrckbp: I reverted the canOutsideClickClose, and i'll see what I can do for the width of the modal

ptrckbp:
I reverted the default width to 500, and added an exception for the node action editor.

EFF (approved):

PR #44, opened 2021-08-05
fix(studio): guided tour highlighting restored and more

ptrckbp:
Guided tour now highlights correct array. It was highlighting nothing for a few nodes and staying in the middle. I removed the bubble, and updated the text to make it slightly less challenging to read. 
image

When right clicking in the flow UI, then clicking on the "skill" submenu button, the menu would close. This is a problem when you move around / click around too fast. 
image
This no longer happens. :)

Finally, there was a soft-crash happening when the user clicked the top intents 
image

It now redirects as it should.

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/GuidedTour.jsx
+++ b/packages/studio-ui/src/web/components/Layout/GuidedTour.jsx
@@ -25,7 +25,7 @@ export default class GuidedTour extends React.Component {
         content: 'Welcome to Botpress! This is a quick tour of the most important features.'
       },
       {
-        selector: '.Pane.Pane1',
+        selector: '#bp-menu_flows',
         content:
           'The Flows screen is the main interface where you can see and edit your conversation flows.'
       },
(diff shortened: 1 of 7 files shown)


packages/studio-ui/src/web/components/Layout/GuidedTour.jsx
```diff
@@ -25,27 +25,31 @@ export default class GuidedTour extends React.Component {
         content: 'Welcome to Botpress! This is a quick tour of the most important features.'
       },
       {
-        selector: '#bp-menu_qna',
-        content: 'The QnA module is great for easily adding knowledge to you bot as "Question & Answer" pairs.'
+        selector: '.Pane.Pane1',
```
allardy, line 28: `#bp-menu_flows`

packages/studio-ui/src/web/components/Layout/Sidebar.tsx
```diff
@@ -96,7 +96,7 @@ const Sidebar: FC<Props> = props => {
 
   return (
     <aside className={classnames(style.sidebar, 'bp-sidebar')}>
-      <a href="admin/" className={classnames(style.logo, 'bp-logo')}>
+      <a href="admin/" className={classnames(style.logo, 'bp-logo')} id="all-bots-navigation">
```
allardy, line 99: I would suggest home or admin

ptrckbp:
I was on the fence about the flow selector to use, but I think you are right about this, at least in the long term. We will likely have more of this in the specific pages. I changed the name too like requested :)

allardy (approved):
LGTM!

PR #45, opened 2021-08-06
fix(studio) content picker confusion

ptrckbp:
I'm not trying to redo the whole content picker because that would take a lot of time. 
This aims only to get rid of the issue I got where I was writing text, then "create new text" appears greyed out, and I subconsciously decided it was created, then clicked x and nothing appeared. :(

The fix lies in adding the "active" state to "create new ${content_type}" when there are no search results, to make it more obvious, and having the "enter" key create or select whatever is active
Screen Shot 2021-08-06 at 2 42 36 PM

This also makes it so clicking the greyed out bar acts as create / update button, not just the icon, because I was clicking that too!
Screen Shot 2021-08-06 at 2 53 04 PM

### DIFF
--- a/packages/studio-ui/src/web/components/Content/Select/Widget.tsx
+++ b/packages/studio-ui/src/web/components/Content/Select/Widget.tsx
@@ -118,7 +118,6 @@ class ContentPickerWidget extends Component<Props> {
       <ControlGroup fill>
         <div
           className={style.clickableInput}
-          style={{ display: 'flex', flexDirection: 'row'}}
           onClick={() => (contentItem ? this.editItem() : window.botpress.pickContent({ contentType }, this.onChange))}
         >
           <InputGroup
--- a/packages/studio-ui/src/web/components/Content/Select/Widget.tsx
+++ b/packages/studio-ui/src/web/components/Content/Select/Widget.tsx
@@ -116,14 +116,19 @@ class ContentPickerWidget extends Component<Props> {
 
     return (
       <ControlGroup fill>
-        <InputGroup
-          placeholder={placeholder}
-          value={textContent}
-          disabled
-          id={inputId || ''}
-          className={style.contentInput}
-        />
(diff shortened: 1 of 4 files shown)


packages/studio-ui/src/web/components/Content/Select/Widget.tsx
```diff
@@ -116,14 +116,19 @@ class ContentPickerWidget extends Component<Props> {
...
-        <InputGroup
-          placeholder={placeholder}
-          value={textContent}
-          disabled
-          id={inputId || ''}
-          className={style.contentInput}
-        />
-        {contentItem && <Button icon="edit" onClick={this.editItem} className={Classes.FIXED} />}
+        <div
+          style={{ display: 'flex', flexDirection: 'row'}}
```
allardy, line 120: Please avoid style when possible, use a className and put the css in the .scss file

packages/studio-ui/src/web/components/Content/Select/index.tsx
```diff
@@ -296,14 +311,17 @@ class SelectContent extends Component<Props, State> {
...
         />
         <hr />
         <div className="list-group">
           {categories.map((category, i) => (
             <a
               key={i}
               onClick={() => this.setState({ newItemCategory: category, newItemData: null })}
-              className={`list-group-item list-group-item-action ${style.createItem}`}
+              className={`list-group-item list-group-item-action ${style.createItem} ${
+                thereIsASearchResult ? '' : i === this.state.activeItemIndex ? 'active' : ''
```
allardy, line 323: Please avoid using ? : twice on the same line. This makes it harder to understand the logic. Instead, you can move them to a named variable and use those

Also, I would suggest a diferent naming: hasSearchResults

ptrckbp:
Code pushed with your suggestions. Let me know if I can add anything else :)

PR #53, opened 2021-08-10
feat(studio): Swapped flow and content buttons for good UX

rohankokkulabp:
Users often tend to navigate to 'content' instead of 'flow' as a default human reflex.
Since "Flow" is our main working entity in studio, swapping these items will improve UX by making first item as "Flows"


rohankokkulabp:
renamed branch

PR #54, opened 2021-08-10
fix: Swapped flow and content menu items in side menu

rohankokkulabp:
Users often tend to navigate to 'content' instead of 'flow' as a default human reflex.
Since "Flow" is our main working entity in studio, swapping these items will improve UX by making first item as "Flows"
sidebar_icon_swap-01


EFF (approved):
Next time please follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0)

allardy:
Please add the Prettier and ESLint extension on VS Code.

When you save a file, it should be formatted automatically. Here, the bracket is not correctly aligned with others.

image

allardy (approved):

PR #67, opened 2021-08-20
fix(studio): shallow copy of node/skill

daukadolt:
closes https://github.com/botpress/botpress/issues/4683


laurentlp:
@daukadolt Can you please add a screen capture that shows the fix in action so it's easier to review?

laurentlp:
Closing this for now. We will re-open when we have a better idea of how we should implement this feature!

PR #69, opened 2021-08-23
chore(studio-be): rm calls to checkForDirtyModels as its already watched by nlu module

franklevasseur:
Hey,

I might be wrong, but I believe there's no need for the studio to manually check for dirty models as nlu module already listen for such files:

If you check [this file](https://github.com/botpress/botpress/blob/master/modules/nlu/src/backend/application/scoped/definitions-service.ts#L61),

You'll see this code:
```ts
return this._definitionRepository.onFileChanged(async filePath => {
  const hasPotentialNLUChange = filePath.includes('/intents/') || filePath.includes('/entities/')
  if (!hasPotentialNLUChange) {
    return
  }
  await Promise.map(this._languages, this._notifyListeners)
})
```

which listens on ghost and notifies to web-socket.

Therefore I don't think studio should have to make a call to `POST /checkForDirtyModels`.

Am I missing something ? Tell me what you think!

### DIFF
--- a/packages/studio-be/src/core/app/core-client.ts
+++ b/packages/studio-be/src/core/app/core-client.ts
@@ -27,9 +27,6 @@ export const coreActions = {
   setStudioReady: async () => {
     await coreClient?.post('/setStudioReady')
   },
-  checkForDirtyModels: async (botId: string) => {
-    await coreClient?.post('/checkForDirtyModels', { botId })
-  },
   syncBotLibs: async (botId: string) => {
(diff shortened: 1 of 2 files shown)


allardy:
onFileChanged is triggered by changes on the ghost-service, which is single-process. 

Since you are changing the intents via the `studio` process, it would trigger onFileChanged on the studio, but the core (which runs the nlu module) wouldn't receive that, so I added that route specifically for that shortcoming

franklevasseur:
understood :+1:

PR #77, opened 2021-08-31
chore(studio): move bot creation to studio

allardy:
Moving the bot creation from the core to the studio is necessary for the next steps (everything contained in the builtin module will be moved here)


packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -197,6 +232,95 @@ export class BotService {
     await this.mountBot(destBotId)
   }
 
+  async getBotTemplates(): Promise<any[]> {
```
laurentlp, line 235: Can you add typings here

packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -197,6 +232,95 @@ export class BotService {
...
+      if (!mergedConfigs.imports.contentTypes) {
+        const allContentTypes = await this.cms.getAllContentTypes()
+        mergedConfigs.imports.contentTypes = allContentTypes.map(x => x.id)
+      }
+
+      if (!mergedConfigs.defaultLanguage) {
+        mergedConfigs.disabled = true
+      }
+
+      await scopedGhost.upsertFile('/', BOT_CONFIG_FILENAME, stringify(mergedConfigs))
```
laurentlp, line 299: On Botpress, we used to also create the folder structure like so:

```js
await scopedGhost.ensureDirs('/', BOT_DIRECTORIES)
```

Shouldn't we do the same here?

packages/studio-be/src/core/misc/archive.ts
```diff
@@ -1,6 +1,60 @@
+import fse from 'fs-extra'
```
laurentlp, line 1: Can you add at least unit tests for the two files in the misc folder.

packages/studio-be/src/core/misc/archive.ts
```diff
@@ -1,6 +1,60 @@
...
+import stream from 'stream'
 import tar from 'tar'
+import tmp from 'tmp'
+import unzipper from 'unzipper'
 import { VError } from 'verror'
 
+import { forceForwardSlashes } from './utils'
+
+// Source: https://github.com/kevva/is-zip
+const isZip = buf => {
```
laurentlp, line 14:
```suggestion
const isZip = (buf: Buffer) => {
```

packages/studio-be/src/core/misc/archive.ts
```diff
@@ -1,6 +1,60 @@
+import fse from 'fs-extra'
+import glob from 'glob'
+import mkdirp from 'mkdirp'
+import path from 'path'
+import stream from 'stream'
 import tar from 'tar'
+import tmp from 'tmp'
+import unzipper from 'unzipper'
```
laurentlp, line 8: Can you add all the @types you can so we have types for those libraries

packages/studio-be/src/core/misc/archive.ts
```diff
@@ -1,6 +1,60 @@
...
+    await new Promise((resolve, reject) => {
+      writeStream.on('close', resolve) // emitted by unzipper
+      writeStream.on('end', resolve) // emitted by tar
+      writeStream.on('error', reject)
+    })
+
+    const files = await Promise.fromCallback<string[]>(cb => glob('**/*.*', { cwd: destination }, cb))
+    return files.map(filePath => forceForwardSlashes(filePath))
+  } catch (err) {
+    throw new VError(err, `[Archive] Error extracting archive to "${destination}"`)
```
laurentlp, line 54: There seem to have a typings issue with the err:

Screenshot from 2021-09-09 17-22-38

packages/studio-be/src/studio/manage/manage-router.ts
```diff
@@ -0,0 +1,64 @@
...
+          await this.botService.addBot(bot, req.body.template)
+        }
+
+        if (botLinked) {
+          this.logger.warn(`Bot "${bot.id}" already linked in workspace. See workspaces.json for more details`)
+        } else {
+          await this.workspaceService.addBotRef(bot.id, req.workspace!)
+        }
+
+        res.sendStatus(200)
```
laurentlp, line 58: We used to return the botId here. Maybe it would be a good idea to keep it that way since the id is generated here in the back-end?

packages/studio-be/src/core/users/workspace-service.ts
```diff
@@ -73,6 +66,13 @@ export class WorkspaceService {
...
   }
 
+  async getPipeline(workspaceId: string): Promise<Pipeline | undefined> {
+    const workspaces = await this.getWorkspaces()
+    const workspace = workspaces.find(x => x.id === workspaceId)
+
+    return workspace?.pipeline
+  }
+
   async findWorkspaceName(workspaceId: string): Promise<string> {
```
laurentlp, line 76:
```suggestion
  async findWorkspaceName(workspaceId: string): Promise<string | undefined> {
```

packages/studio-be/src/core/users/workspace-service.ts
```diff
@@ -84,6 +84,24 @@ export class WorkspaceService {
     return list.find(x => x.workspace === workspace)
   }
 
+  async save(workspaces: Workspace[]): Promise<void> {
+    return this.ghost.global().upsertFile('/', 'workspaces.json', JSON.stringify(workspaces, undefined, 2))
+  }
+
+  async addBotRef(botId: string, workspaceId: string): Promise<void> {
+    const workspaces = await this.getWorkspaces()
+    const workspace = workspaces.find(x => x.id === workspaceId)
```
laurentlp, line 93: I think you should move this piece of code into a function since it's copy-pasted at multiple occurrences in this file.

packages/studio-be/src/core/users/workspace-service.ts
```diff
@@ -84,6 +84,24 @@ export class WorkspaceService {
...
+
+    if (!workspace) {
+      throw new Error(`Specified workspace "${workspaceId}" doesn't exist`)
+    }
+
+    if (!workspace.bots.includes(botId)) {
+      workspace.bots.push(botId)
+    }
+    return this.save(workspaces)
```
laurentlp, line 102: You should only save the workspace if it was updated otherwise it calls the ghost for nothing.

packages/studio-be/package.json
```diff
@@ -9,7 +9,7 @@
     "start": "cd ./out && cross-env NODE_PATH=./ cross-env BP_MODULES_PATH=B:/bp18/modules node index.js",
     "build": "yarn clean && yarn tsc",
     "clean": "node ./node_modules/rimraf/bin.js ./out",
-    "watch": "yarn clean && yarn tsc --watch"
```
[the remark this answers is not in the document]
allardy: Because it should be a single-purpose command. And everytime I want to use yarn watch, it's when I'm working in the code, it always fails because my server is still running, so I have to kill the server, type yarn watch, then start it again. Better dx

packages/studio-be/src/core/bots/bot-service.ts
allardy: Not necessary, I tested it, removed that part

packages/studio-be/src/core/misc/archive.ts
allardy: Not touching archive, it was copied from the main repo. Added a comment

packages/studio-be/src/core/misc/archive.ts
allardy: This is related to a new typescript version in vscode, unrelated

packages/studio-be/src/studio/manage/manage-router.ts
allardy: Good point

packages/studio-be/src/core/users/workspace-service.ts
allardy: Not necessary, it throws if not found

packages/studio-be/src/core/users/workspace-service.ts
allardy: Necessary here because we need to save all workspaces, not just the one we need. Else we have findWorkspace

PR #81, opened 2021-08-31
fix(cms): display unregistered content-types instead of error message

laurentlp:
This PR adds support for editing content-elements event when at least one of the bot's content-types are unregistered. It also displays the list of unregistered content-types.

### Before

When you had an unregistered content-type, you would get this error on the CMS page:

Screenshot from 2021-09-03 12-21-29

### Now

Screenshot from 2021-09-03 12-36-20

Screenshot from 2021-09-03 12-36-35

Screenshot from 2021-09-03 13-25-46

Closes DEV-1519


sadda11asm (approved):
LGTM! Thanks :)

PR #86, opened 2021-09-03
fix(studio): bot mig same behavior as core

allardy:
Bot migrations were not taking into account TESTMIG_NEW or TESTMIG_ALL, so it was not in sync with the core. Simple fix, copied the logic from core.


laurentlp:
Does it fix this issue (https://github.com/botpress/botpress/issues/5409)?

laurentlp (dismissed):
LGTM!

allardy:
It didn't fix the issue correctly so I did a minor change. It worked when importing a bot, but not when starting the server

laurentlp (approved):

PR #90, opened 2021-09-08
fix(studio-ui): remove id from bot form parameters

billsomen:
**Descriptiion:**
When saving the bot configuration, there is an error 
(image below)
image

The bot id is a readonly input and I just removed it from the bot error config...
now we have 
image


allardy (approved):
LGTM

PR #105, opened 2021-09-15
fix(studio): retry set studio ready call

samuelmasse:
I was able to reproduce the bug where the studio would be ready before the runtime server by placing a Promise.delay somewhere. This fix prevents the studio from crashing with a ECONNREFUSED error and just keeps retrying to inform the runtime that it is ready.

The fix works, but I'm not really sure why we start the studio before the runtime server is started. Maybe a better fix in the future would be to boot the runtime completely and then start the studio

Closes DEV-1470


allardy (approved):
It will fix that issue, it's mostly temporary until we remove the studio entirely... should do the trick.

PR #110, opened 2021-09-21
chore(nlu): move nlu module code in studio

franklevasseur:
The nlu module has 2 responsibilities: training and prediction.

This PR takes pretty much all the training part of the nlu module and moves it in the studio-be codebase.


packages/studio-be/src/core/app/bootstrap.ts
```diff
@@ -99,7 +99,7 @@ async function start() {
   const modules = _.uniqBy(globalConfig.modules, x => x.location)
   const enabledModules = modules.filter(m => m.enabled)
 
-  const logger = await getLogger(app.logger, 'Launcher')
+  const logger = await getLogger(app.logger, '[Studio] Launcher')
```
allardy, line 102: All logs of the studio are already prepended by [Studio], this can be removed

packages/studio-be/src/core/app/inversify/app.inversify.ts
```diff
@@ -74,6 +74,11 @@ container
   .to(ConfigProvider)
   .inSingletonScope()
 
+container
+  .bind<MonitoringService>(TYPES.MonitoringService)
+  .to(CEMonitoringService)
+  .inSingletonScope()
```
allardy, line 80: The monitoring service has absolutely nothing to do with the studio, it should stay in the core

packages/studio-be/src/core/health/alerting-service.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+@injectable()
+export class CEAlertingService implements AlertingService {
+  async getIncidents(_dateFrom: number, _dateTo: number): Promise<Incidents | undefined> {
+    return
+  }
+  start(): void {}
+  stop(): void {}
+}
```
allardy, line 40: This has nothing to do with the studio

packages/studio-be/src/studio/nlu/application/model-state/model-state-repo.ts
```diff
@@ -0,0 +1,92 @@
...
+      table.string('botId').notNullable()
+      table.string('language').notNullable()
+      table.string('status_type').notNullable()
+      table.string('modelId').notNullable()
+      table.string('definitionHash').notNullable()
+      table.string('status').notNullable()
+      table.float('progress').notNullable()
+      table.primary(['botId', 'language', 'status_type'])
+    })
+  }
```
allardy, line 44: The studio is meant to be executed on a user's machine and it shouldn't require access to the database. The only reason we keep the database logic in the studio backend is so we can ship it right now and it work seamlessly for users in production.

It should certainly not create a database schema

packages/studio-be/src/studio/nlu/application/non-blocking-app.ts
```diff
@@ -0,0 +1,40 @@
...
+        const health = await this.getHealth({ reportError: false })
+        if (health?.isEnabled) {
+          clearInterval(i)
+          resolve()
+          return
+        }
+      }, STAN_POLLING_INTERVAL)
+    })
+  }
+}
```
allardy, line 40: I don't think this logic is necessary. The studio should not wait after anything. When you press "Train now", it tries to contact the nlu server, if it's not up, just display an error message. We should stop making processes waiting for each other, it just adds complications

packages/studio-be/src/studio/nlu/non-blocking-app.ts
```diff
@@ -0,0 +1,40 @@
...
+        const health = await this.getHealth({ reportError: false })
+        if (health?.isEnabled) {
+          clearInterval(i)
+          resolve()
+          return
+        }
+      }, STAN_POLLING_INTERVAL)
+    })
+  }
+}
```
allardy, line 40: This file is there twice

packages/studio-be/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "paths": {
       "botpress/sdk": ["./sdk/botpress.d.ts"]
     },
-    "types": ["reflect-metadata", "jest"]
+    "types": ["reflect-metadata", "jest"],
+    "skipLibCheck": true
```
allardy, line 20: Why ?

packages/studio-ui/src/web/components/Layout/LangServerHealthWarning.tsx
```diff
@@ -6,7 +6,7 @@ import React, { useEffect } from 'react'
 export default () => {
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    axios.get(`${window.BOT_API_PATH}/mod/nlu/health`).then(({ data }) => {
+    axios.get(`${window.STUDIO_API_PATH}/nlu/health`).then(({ data }) => {
```
allardy, line 9: Shouldn't the health be on the runtime? The studio only trains a model. It shouldn't care about its health

allardy (changes requested):
We are currently trying to split the various processes, why make the studio go through the core as a proxy ? Currently the studio reads the botpress.config.json file by itself, so please make sure that the Studio backend communicates with the NLU server directly.  

The studio should handle everything related to training.
The runtime should handle the logic for prediction.

The only communication I see is between the studio and the core so it can push its realtime updates through the socket.

You can easily start the NLU server, then start the studio, providing the necessary information for it to be used. The studio *should not* depend on anything, it should be feeded everything

Also, please remove the logic related to the database, The studio backend should only use files (the database support is temporary just so it's not breaking)

packages/studio-be/src/core/app/bootstrap.ts
franklevasseur: yep leftover

packages/studio-be/src/core/app/inversify/app.inversify.ts
franklevasseur: yep, leftover, I tried bringing back the `RealtimeService` in studio

I forgot to remove

packages/studio-be/src/core/health/alerting-service.ts
franklevasseur: yep same, still a leftover

packages/studio-be/src/studio/nlu/application/non-blocking-app.ts
franklevasseur: The train now button can't be displayed If we don't know if a bot needs a training. To know if a bot needs a training, we have to query the nlu-server.

This is a long and complicated topic, but long story short: I submitted a new UI to the product team so that the studio UI would be a little more representative of the nlu-server API, but my suggestion was rejected. So yeah, we have to keep that.

packages/studio-be/tsconfig.json
franklevasseur: typescript was trying to build an npm package

packages/studio-ui/src/web/components/Layout/LangServerHealthWarning.tsx
franklevasseur: I didnt think about that 🤔

This could work better, I have to think about it

packages/studio-be/src/studio/nlu/application/model-state/model-state-repo.ts
franklevasseur: This information needs to be shared between different botpress instances... If not it won't work.

It's either in the kvs, the database or the ghost. I just thought the database was better suited

packages/studio-be/tsconfig.json
franklevasseur: I removed it and it builds without problem

packages/studio-be/src/studio/nlu/application/model-state/model-state-repo.ts
allardy: I suggest using a file on the ghost to keep track of the progress. When the studio will run as a standalone, the user will build his bot locally, train it, and export the .tgz archive which will contain the trained model. Building it this way is future-proof for when we will stop shipping the studio with the rest of the product.

The ghost is future-proof, since when we will just ship the studio as  a standalone, we can simply remove the database logic and it will use the disk driver only

packages/studio-be/src/studio/nlu/application/non-blocking-app.ts
allardy: When you open the studio, you can make a query to the NLU server to check if training is required. I don't understand why the studio has to wait for the nlu server to be ready

franklevasseur: This is non-blocking for the studio, but anyway, I think you're right. 

nlu-server should be up when studio starts, if not we'll log an error.

PR #115, opened 2021-09-23
fix(qna): fix language dropdown overlay issue 

EFF:
The issue is easily reproductible. I thought it was a z-index thing but turns out z-index don't apply to `position:relative` elements. Removing that position relative statement had me play around with a bunch of styles. In the end, there is no visible change except the actual fix (see screencaps)

Before  :
Screen Shot 2021-09-23 at 11 48 16 AM
 
After:
Screen Shot 2021-09-23 at 11 46 59 AM

Closes https://github.com/botpress/botpress/issues/5417 DEV-1693


allardy (approved):
LGTM

PR #116, opened 2021-09-23
feat(studio): implement realtime between be/fe

allardy:
Quick implementation of the realtime service between the backend and frontend, we will need it eventually (not sure of the behavior since the studio backend is on a different port, must try it)


EFF:
What's that for ? we will need this for what ? do we keep this open ? @allardy

allardy:
Will be used in next when the studio will run alone, now the studio frontend communicate with the botpress backend, and not the studio backend. on hold

PR #118, opened 2021-09-23
fix(studio): prevent creating empty actions

### LINKED ISSUE #76: [BUG] It's possible to add an empty action
**Describe the bug**
image
You can click add action without anything

**To Reproduce**

1. Go to flow
2. Create a node
3. Click "+" sign then add action
4. See action created

**Expected behavior**
Form validation should catch this. The Add action button should be disabled unless there is something to add.

**Environment (please complete the following information):**

- OS: macOS
- Browser Chrome
- Browser Version Version 92.0.4515.131 (Official Build) (x86_64)
- Botpress Version 12.25.0
- NodeJS 12.13.0

**Additional context**
This will break the flow at the added location. Since it's an add function rather than an update function, I think in most cases people should notice it.

laurentlp:
This PR fixes an issue where it was possible to add empty actions to a node. To fix this issue, a validation was added to make sure the form is valid before enabling the submit button.

**Before**

https://user-images.githubusercontent.com/9640576/134557976-7e5ce35f-d8a4-416d-8d86-5a126bbf86d6.mp4

**After**

https://user-images.githubusercontent.com/9640576/134557986-284807fd-f5d8-41d4-a882-4cf1e196edaf.mp4

Closes #76 and DEV-1638

### DIFF
--- a/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ActionModalForm.tsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ActionModalForm.tsx
@@ -20,7 +20,7 @@ interface Action {
   metadata: LocalActionDefinition
 }
 interface Item {
-  type: string
+  type: ActionType
   functionName?: string
   message?: string
   parameters: Parameter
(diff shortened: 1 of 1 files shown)


EFF (approved):
simple enough. LGTM.

You can wait for @allardy's input but I believe this is fine ! 

What happens with skills ? does it change anything

Issue #108, opened 2021-09-17
[BUG] Content of carousel in arabic is not saved


Jeremynano:
I have FR/AR bot.

I'am trying to translate the content of my carousel from FR to AR.

- I set my bot to AR
- 1) I click on the title or content of my carousel
- 2) I write in arabic
- 3) I click outside the input
- 3 bis) The new arabic content disappear and the old content (in french) appear again

![image](https://user-images.githubusercontent.com/27802887/133796071-ff4884dc-5495-4eef-88c5-1dcb0bd55fd6.png)

Environment (please complete the following information):

OS: Windows
Browser Chrome Version 92.0.4515.159 / Firefox Version 91.0.2
Version : 12.26.2

PR #122, opened 2021-09-24
fix: edit translated caroussel data

EFF:
Notice this line is the original code, the comment mentions we should ignore array arrays of empty objects
https://github.com/botpress/studio/blob/fed7ac66da61b6ac3a5cc0d43ca1106806934e7c/packages/studio-ui/src/web/util/NodeFormData.ts#L18-L34

In this code at line 27, we return undefined if the array has **not empty** objects `!formData[x].filter(_.isEmpty).length`

I rewrote the logic for something much simpler, yet readable.

Fixes https://github.com/botpress/studio/issues/108
Closes : DEV-1825 & DEV-1086


laurentlp:
@EFF What's the link between this PR and DEV-1580?

packages/studio-ui/src/web/util/NodeFormData.ts
```diff
@@ -1,36 +1,24 @@
 import _ from 'lodash'
 
 export const getFormData = (node: any, contentLang: string, defaultLanguage: string, defaultValue: any = {}): any => {
```
laurentlp, line 3: Can you add missing typings!

laurentlp (commented):
Small comment other than that LGTM!

laurentlp (dismissed):
Tested and working fine! Good job!!

EFF:
@laurentlp I added typings but it's the worst typing we could add as by debugging, there is a single key in the object and its any. We also read a second key that is also a type any and return it as is if it's defined (although I didn't manage to see it defined).

To be honest that whole file should be re-written properly

laurentlp (approved):

PR #125, opened 2021-09-29
chore: few fixes and changes on branch next

franklevasseur:
**WIP**

### What's left on the nlu side

- [ ] rewrite model state repo so it uses the ghost instead of database
- [x] make sure studio does not "**wait**" for nlu-server to be up before booting


franklevasseur:
I'm merging "as is" even if the model state repo is still using the database instead of the ghost. This really needs to be merge and we can still do a migration the day we decide to drop the database

PR #127, opened 2021-09-30
fix(flow): fix ctrl + s toast text display

### LINKED ISSUE #99: [Bug] Ctrl + S on flows shows translation keys instead of the value
In the studio, go to the "Flows" tab. Press "Ctrl + S" it used to show a message that there is no need to save as now it is auto saved. Now it doesn't

Botpress v12.26.1

image

laurentlp:
This PR fixes a translation when pressing Ctrl + S on the flow editor

**Before**

Screenshot from 2021-09-30 11-34-44

**After**

Screenshot from 2021-09-30 11-34-22

Closes #99
And closes DEV-1705

### DIFF
--- a/packages/studio-ui/src/web/views/FlowBuilder/index.tsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/index.tsx
@@ -136,7 +136,7 @@ const FlowBuilder = (props: Props) => {
     },
     save: e => {
       e.preventDefault()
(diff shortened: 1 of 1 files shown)


tomkcey (approved):
Just pulled the `studio` repository and that's where the string is stored, plus your screenshot shows it's working so I'm fine with this.

PR #129, opened 2021-09-30
fix(studio): added form validation to bot settings page

laurentlp:
This PR fixes an issue where it was possible to set a bot name with more than 50 chars which isn´t valid and throws a validation error in the front-end

Related to: https://github.com/botpress/botpress/pull/5524

### DIFF
--- a/packages/studio-ui/src/web/views/Config/index.tsx
+++ b/packages/studio-ui/src/web/views/Config/index.tsx
@@ -344,7 +344,13 @@ class ConfigView extends Component<Props, State> {
                   </div>
                 </FormGroup>
                 <FormGroup label={lang.tr('name')} labelFor="name">
-                  <InputGroup id="name" name="name" value={this.state.name} onChange={this.handleInputChanged} />
+                  <InputGroup
+                    id="name"
+                    name="name"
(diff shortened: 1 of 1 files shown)


tomkcey (commented):
Looking good to me. The user won't be notified that the max length is 50 characters though. Just making sure we have a way to warn him/her?

laurentlp:
> Looking good to me. The user won't be notified that the max length is 50 characters though. Just making sure we have a way to warn him/her?

The user simply won't be able to type more than 50 chars in the input field. That's what the `maxLength` propertiy does to the component.

tomkcey:
> > Looking good to me. The user won't be notified that the max length is 50 characters though. Just making sure we have a way to warn him/her?
> 
> The user simply won't be able to type more than 50 chars in the input field. That's what the `maxLength` propertiy does to the component.

I know 😄 My question was is this what we want in terms of UX (User Experience). Do we have a warning component that will show the user he<s currently doing something that is forbidden, or will it silently fail? I don<t think the native HTML input component does any warning does it?

allardy (approved):
Let's keep it simple! LGTM

PR #134, opened 2021-10-06
fix(ui): modify channel-vonage content-type warn

charlescatta:
Small UI change, pretty self-explanatory.

Closes https://github.com/botpress/v12/issues/1441
closes: DEV-1568

Changed warning sign:
Screenshot from 2021-10-06 17-35-04

Screenshot from 2021-10-06 17-35-48

### DIFF
--- a/packages/studio-ui/src/web/components/ContentForm/index.tsx
+++ b/packages/studio-ui/src/web/components/ContentForm/index.tsx
@@ -104,17 +104,15 @@ const CustomDescriptionField = ({ description, id, formContext }: FieldProps) =>
   }
 
   if (id === 'root__description' && Object.keys(mapping).includes(formContext.subtype)) {
-    const capitalize = (str: string) => {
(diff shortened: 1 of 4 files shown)


laurentlp (changes requested):
We decided to keep the message and go with this warning instead: `Please note that <CONTENT_TPE> content-type is only supported in channel-web and channel-vonage` (See: https://github.com/botpress/botpress/issues/5353)

Also if we wanted to be more precise, we could do a mapping like so:

```
const mapping = {
  audio: [ 'channel-web', 'channel-vonage' ],
  video: [ 'channel-web', 'channel-vonage' ],
  file: [ 'channel-vonage' ],
  location: [ 'channel-web', 'channel-vonage' ]
}
```

and use this mapping when displaying the warning

EFF (dismissed):

EFF (approved):

PR #147, opened 2021-10-25
feat: adds Segment integration

ptrckbp:
This adds Segment to the Studio for those who have analytics enabled. 

Apart from the basic loading of the library, there are three pieces of metadata we are collecting. All three are hashed and collected with the [Segment identify function](https://segment.com/docs/connections/spec/identify/). 

accountLicenseHash
userEmailHash
machineUUID

This should give us a good idea of how many people are using Botpress in a given team. 

I'm not entirely sure about whether this follows the application logic enough, but comment away. I'll change it according to what works best with you guys, and then submit a PR for the admin panel too. :) 

The userId is unset on purpose, and will also be unset in the Admin panel, which will mean it will be shared.


allardy (approved):
LGTM

PR #156, opened 2021-11-01
fix: CallAPI Transition Edit Not Persisted

Ierezell:
Linear issue: DEV-1260

This fixes https://github.com/botpress/botpress/issues/4571 

https://www.awesomescreenshot.com/video/5854955?key=8bdc26c0b9070f9c41ed66f19e42f1a9

Fix is to save a new field in the `flow.json` named `conditionType` instead of inferring it at load time. 

New `Flow.json` : 
image

Note: The one transition added with this new behavior has a `"conditionType": "raw"` field. 
It is not breaking and backward compatible. 
 

Note 2 : 
This does not migrate actions! 

The old behavior was : "Create a raw action" => "Type is badly inferred to MatchedProps" 
New behavior: "Create a raw action" => "Save it and load it as raw action"

New behavior for old actions "Load it: as there is no `conditionType` infer type (e.g load a MatchedProps), save it as MatchedProps". If the user wanted to change it to a raw action he need to do it by hand as we cannot magically know the real type of each action.

### DIFF
--- a/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ConditionModalForm.jsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ConditionModalForm.jsx
@@ -47,7 +47,7 @@ class ConditionModalForm extends Component {
 
     const { item } = this.props
     const condition = (item && item.condition) || ''
-    const conditionType = this.getConditionType(condition)
+    const conditionType = (item && item.conditionType) || this.getConditionType(condition)
 
     if (item && item.node) {
       let typeOfTransition = item.node.indexOf('.') !== -1 ? 'subflow' : 'node'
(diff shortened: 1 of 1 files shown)


laurentlp (approved):

crochcunill:
Solved in version 12.26.7

crochcunill:
Actually I am finding the same problem. It takes some patience because most of the time seems the information persist, but then, you return and realize the condition for the transition has changed. 

In the last case I altered the header tab in the API skill (fill it with {}), after this change  it showed 
image
The Condition was set the "Raw Expression" and has changed to "Matches property" to a temporary dialog box  **valid_6drvakawdp**. I have no idea where these values comes from


I have attached a copy of the bot. Please change the extension from txt to tgz, as github does not allow to upload tgz files.
[bot_apitest_1638405010227.txt](https://github.com/botpress/studio/files/7637976/bot_apitest_1638405010227.txt)


Using BP version 12.26.7
Chrome, Mac

PR #157, opened 2021-11-01
fix(studio): prevent using skills in folde rname

allardy:
image

Fixes https://linear.app/botpress/issue/DEV-1425/[bug]-flow-editor-skills-folder-is-nowhere-botpressstudio-43


sebburon (approved):
Looking good

Issue #52, opened 2021-08-09
[BUG] renaming to existing node name soft-fails


ptrckbp:
![Peek 2021-07-19 09-25](https://user-images.githubusercontent.com/87815239/128723960-d3c00590-5dc9-4059-8be5-8cc2f7eed11b.gif)

There should be an error message if this happens.

PR #159, opened 2021-11-01
fix(studio): fix error when renaming node

sebburon:
fixes https://github.com/botpress/studio/issues/52

Added a toast at the bottom of the screen.
Screen Shot 2021-11-01 at 11 29 18 AM


allardy:
I'm getting a weird bug. Create a new flow, then add a new node on that flow.
Click on the node to open the inspector. Then click outside the node (without changing anything), and you will get the error toast. Add other nodes, it will repeat the same issue. Switching a node and clicking on the background triggers the toast

sebburon:
> I'm getting a weird bug. Create a new flow, then add a new node on that flow.
> Click on the node to open the inspector. Then click outside the node (without changing anything), and you will get the error toast. Add other nodes, it will repeat the same issue. Switching a node and clicking on the background triggers the toast

The node is being renamed every time we inspect it...
Working on it !

packages/studio-ui/src/web/views/FlowBuilder/nodeProps/StandardNode.jsx
```diff
@@ -12,9 +12,11 @@ const style = require('./style.scss')
 
 export default class StandardNodePropertiesPanel extends Component {
   renameNode = text => {
-    if (text) {
+    if (text !== this.props.node.name) {
```
allardy, line 15: This brings another issue. An empty name will create a node where all the other nodes will point to
image

Also, please rename the method, "renameNode" is not accurate, it's "onChange". So we need to check if text is set and is not blank

sebburon: You are right my bad !

allardy (approved):
LGTM

PR #160, opened 2021-11-01
fix(studio): no ghost flows when deleting a bot

allardy:
When you delete a bot, flows in the studio are not cleared correctly so they stayed in memory and were displayed if you re-created a bot with the same id.

This is a quick workaround to cleanup correctly 

Fixes https://linear.app/botpress/issue/DEV-1670/[bug]-deleted-bots-not-being-deleted-properly-botpressbotpress-5393****


ptrckbp (approved):
Quite simple, and does the trick!

PR #168, opened 2021-11-01
fix(studio): warns user about double braces

ptrckbp:
image
(the warning message is added)

Really simple. Doesn't attempt to standardize everything, but instead warns user when content requires triple braces. 

The use case is when a user gets an image / file / video link from somewhere, and needs to display it.


allardy (approved):
LGTM

franklevasseur:
Closes: https://linear.app/botpress/issue/DEV-1637/%5Bbug%5D-inconsistent-escaping-behavior-botpressbotpress-5370

PR #170, opened 2021-11-01
fix(studio): can't lose the bottom debugger panel because of window resize anymore

slvnperron:
If you resized the debugger panel then came back in a smaller window size, the panel would be out of the screen and impossible to get back without clearing your localStorage.

This happened to me all the time...

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/index.tsx
+++ b/packages/studio-ui/src/web/components/Layout/index.tsx
@@ -189,6 +189,15 @@ const Layout: FC<Props> = (props: Props) => {
   const bottomPanelHeight = props.bottomPanelExpanded ? EXPANDED_PANEL_HEIGHT : lastSize
   const bottomBarSize = props.bottomPanel ? bottomPanelHeight : '100%'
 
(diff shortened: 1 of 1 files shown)


crochcunill:
Not completely fixed with version **botpress-prev-darwin**
It is working  when opening the Bottom panel from the studio.
When opening the bottom panel from admin, then the bottom panel is gone and can not be recovered

**Test Steps**

1. Have the screen maximized
2. Log to BP 
3. In the admin page, click on the Bottom Panel icon and verify the Bottom Panel appear in the bottom of the page
4. Place the cursor on the boundary between the two panels to resize their sizes
5. Resize to make the bottom panel small
6. Then reduce the size of the window, so you can not see the bottom panel
7. Then click the Bottom Panel icon (from step 1) , it makes the bottom panel disappear (although you can not see it going away)
8. Click the Bottom Panel icon again

**Actual**
I can see no trace of the bottom panel

**Expected**
Doing the same steps from Studio, I can see a sliver of the bottom panel, that at least indicates it exists

PR #172, opened 2021-11-01
chore(debugger): fix missing translation

laurentlp:
This PR fixes a missing translation in the debugger

Fixes https://github.com/botpress/studio/issues/169
Closes DEV-1961

**Before**

Screen Shot 2021-09-28 at 2 39 26 PM

**After**

Screenshot from 2021-11-01 16-07-27

PR #173, opened 2021-11-01
fix(studio): missing translation in debugger panel

slvnperron:
There was a missing translation in the debugger that didn't look professional and bugged me for a while.

## Before

image

## After

image


laurentlp:
@slvnperron Already done here: https://github.com/botpress/studio/pull/172

Issue #177, opened 2021-11-01
[BUG]: Inspector missing translations


franklevasseur:
This is unacceptable:

![image](https://user-images.githubusercontent.com/25970722/139742798-0f76d818-6ddd-4b1d-8eac-fe28b2aeeb25.png)

PR #180, opened 2021-11-01
fix(studio): added translations in node and flow inspector

franklevasseur:
Fixes: https://github.com/botpress/studio/issues/177
Closes: DEV-1965

looks so nice now:
image


charlescatta:
LGTM totally an 8 point PR, incredibly terrific work. This would have taken me a full week for sure.
Thanks for fixing your e2e tests!

Best regards.

PR #182, opened 2021-11-01
feat(studio): automatic node layout

michaelmass:
## Feature

A new button was added to automatically layout the current flow nodes.

Animation


packages/studio-ui/src/web/views/FlowBuilder/diagram/index.tsx
```diff
@@ -141,6 +142,8 @@ class Diagram extends Component<Props> {
       zoomToLevel: this.props.zoomToLevel
     })
 
+    setInterval(() => this.props.refreshFlowsLinks(), 1000)
```
michaelmass, line 145: I know this needs to change before merging but I didn't find the solution to refresh the diagram. Maybe someone knows how to do it properly.

laurentlp: @allardy? Personally, I'm not sure how it can be done and honestly, the redux code of the flow editor is kind of a mess but we could probably "listen" to changes in the flow `currentHashes` and then update the links or pass some kind of props to your components so that it can trigger a function that handles correctly updating the links upon update.

packages/studio-ui/package.json
```diff
@@ -53,7 +54,7 @@
     "moment": "^2.29.1",
     "ms": "^2.1.3",
     "mustache": "^2.3.0",
-    "nanoid": "^2.0.3",
+    "nanoid": "^3.1.30",
```
laurentlp, line 57: Can you make sure to follow [this upgrade guide](https://github.com/ai/nanoid/releases/tag/3.0.0) or stay on v2.

packages/studio-ui/src/web/views/FlowBuilder/diagram/LayoutToolbar/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+const flowToElkGraph = (flow: FlowView): ElkNode => ({
+  id: 'root',
+  children: flow.nodes.map(value => ({
+    id: value.id,
+    layoutOptions: {},
+    width: 250,
+    height: 200
```
laurentlp, line 40: Where do those values come from?

laurentlp (commented):
Small comments and also, there seem to be a warning print by the ELK constructor:

Screenshot from 2021-11-04 10-31-00

packages/studio-ui/src/web/views/FlowBuilder/diagram/index.tsx
allardy: Why don't you pass the refreshFlowLinks to the toolbar so you can call it once youre done?

allardy (commented):
Be careful with the nanoid change. This breaks the session reset, the logs panel and the skill slot

packages/studio-ui/src/web/views/FlowBuilder/diagram/index.tsx
laurentlp: I quickly tried it and calling `refreshFlowsLinks` directly after the flow update does not work. Maybe it's because the update is done async and there is no way, in the layout components, to know that it is done updating!

laurentlp:
Closing for inactivity. We will re-open this later on.

PR #188, opened 2021-11-01
feat(studio): diagram option to fit the current flow

michaelmass:
## Feature
A new button to fit the current flow into the screen.

Animation


packages/studio-ui/src/web/reducers/ui.ts
```diff
@@ -92,6 +93,12 @@ const reducer = handleActions(
         zoomLevel: state.zoomLevel + 25
       }
     },
+    [zoomToFit]: (state, {}) => {
+      return {
+        ...state,
+        zoomLevel: -1
```
laurentlp, line 99: Can you put this into a reusable constant so we know that this magic number means that the zoom should be adjusted?

laurentlp (commented):
Simple enough! Nice work!!!

I have a small comment otherwise LGTM!

allardy:
Please add translations for the buttons in the translation files, eg `en.json`

laurentlp:
> Please add translations for the buttons in the translation files, eg `en.json`

Right, good catch, I forgot to mention it!

EFF:
@laurentlp @allardy  I fixed your comments

Issue #140, opened 2021-10-18
[BUG] Can't edit image in carousel


Jeremynano:
**Describe the bug**
Can't edit image in carousel

**To Reproduce**

1. Go to "content" -> "carousel -> "i edit my carousel"
2. Click on image of the carousel
3. Nothing happen, can't edit the current image nor change it to fit the second langage i use (FR / AR)

**Expected behavior**
- Make it possible to edit image in carousel
- Make it possible to set different image in carousel based on language 

**Environment (please complete the following information):**

- OS: Windows
- Browser : all browser
- Browser Version all recent version of chrome / edge / mozilla
- Botpress Version  12.26.5

**Additional context**
I'am using AR/FR languages / No errors in console

PR #189, opened 2021-11-02
fix(content): fix carousel image delete button

charlescatta:
Tested, works now.

Caused by onChange handler disallowing an empty value in validation

Fixes https://github.com/botpress/studio/issues/140

### DIFF
--- a/packages/studio-ui/src/web/components/ContentForm/UploadWidget/index.tsx
+++ b/packages/studio-ui/src/web/components/ContentForm/UploadWidget/index.tsx
@@ -39,9 +39,9 @@ const UploadWidget: FC<IUploadWidgetProps> = props => {
     setError(error)
   }
 
-  const onChange = (value: string | null) => {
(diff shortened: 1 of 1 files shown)


packages/studio-ui/src/web/components/ContentForm/UploadWidget/index.tsx
```diff
@@ -39,9 +39,9 @@ const UploadWidget: FC<IUploadWidgetProps> = props => {
     setError(error)
   }
 
-  const onChange = (value: string | null) => {
+  const onChange = (value: string | null, allowEmpty: boolean = false) => {
     const error = validate(value)
```
laurentlp, line 43: PP: It could be simplified like so. I mean you don't need to call validate when allowEmpty is set to true. But since validate is really simple you can keep it like that.

```suggestion
    if (!allowEmpty) {
      const error = validate(value)
    }
    // ...
```

laurentlp (approved):
Simple enough1 Tested and LGTM!

PR #192, opened 2021-11-04
fix(studio): more time for logging that user is logged out

franklevasseur:
Small UI fixes for https://github.com/botpress/studio/pull/185


ptrckbp (approved):
All good, verified in a call.

PR #199, opened 2021-11-11
chore(dev): new migration mechanism

allardy:
The migration mechanism is wobbly and unreliable. It relies on versions stored in config files or in the database which sometimes are not in sync, it's complicated to test new migrations. Before implementing this on the main repository, i'm adding it to the studio since there is way less migrations and the process is already a mess, anyway.

When you mount a bot, it will try to run all migrations in a "dry run" state. In that mode, they don't save any changes, but they return a true/false value if they were about to save something.

Then, we can display the migration label and list migrations that needs to be executed in a compact and more concise way, ex:
image

Previously, it would display this:
image

### DIFF
--- a/packages/studio-be/package.json
+++ b/packages/studio-be/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "start": "cd ./out && cross-env NODE_PATH=./ cross-env BP_MODULES_PATH=B:/bp18/modules node index.js",
-    "build": "yarn clean && yarn tsc",
+    "build": "yarn tsc",
     "clean": "node ./node_modules/rimraf/bin.js ./out",
     "watch": "yarn tsc --watch"
   },
--- a/packages/studio-be/src/core/app/botpress.ts
+++ b/packages/studio-be/src/core/app/botpress.ts
@@ -137,8 +137,7 @@ export class Botpress {
       }`
     )
 
-    const maxConcurrentMount = parseInt(process.env.MAX_CONCURRENT_MOUNT || '5')
-    await Promise.map(botsToMount, botId => this.botService.mountBot(botId), { concurrency: maxConcurrentMount })
+    await Promise.map(botsToMount, botId => this.botService.mountBot(botId))
   }
 
   private async initializeServices() {
--- a/packages/studio-be/src/core/migration/bot-migration-service.ts
+++ b/packages/studio-be/src/core/migration/bot-migration-service.ts
@@ -20,19 +20,10 @@ export class BotMigrationService {
     private bpfs: GhostService
   ) {}
 
-  async executeMissingBotMigrations(botId: string, currentVersion: string, isDown?: boolean) {
-    debug.forBot(botId, 'Checking missing migrations for bot ', { botId, currentVersion, isDown })
+  async executeMissingBotMigrations(botId: string, currentVersion: string) {
+    debug.forBot(botId, 'Checking missing migrations for bot ', { botId, currentVersion })
 
-    if (process.env.TESTMIG_NEW) {
-      const isBotOlder = semver.lt(currentVersion, process.BOTPRESS_VERSION)
-      currentVersion = isBotOlder ? currentVersion : process.BOTPRESS_VERSION
-    }
-
-    const missingMigrations = this.migService.filterMigrations(this.migService.getAllMigrations(), currentVersion, {
-      isDown,
-      target: 'bot'
-    })
-
+    const missingMigrations = await this.getMissingMigrations(botId, this.migService.getAllMigrations())
     if (!missingMigrations.length) {
       return
     }
@@ -43,8 +34,7 @@ export class BotMigrationService {
     }, botId)
 
     try {
-      this.displayMigrationStatus(currentVersion, missingMigrations, this.logger.forBot(botId))
-      await this.executeBotMigrations(botId, missingMigrations)
+      await this.executeBotMigrations(botId, missingMigrations, currentVersion)
     } finally {
       captureLogger.dispose()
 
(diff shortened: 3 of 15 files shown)


packages/studio-be/src/core/app/botpress.ts
```diff
@@ -137,8 +137,7 @@ export class Botpress {
       }`
     )
 
-    const maxConcurrentMount = parseInt(process.env.MAX_CONCURRENT_MOUNT || '5')
-    await Promise.map(botsToMount, botId => this.botService.mountBot(botId), { concurrency: maxConcurrentMount })
+    await Promise.map(botsToMount, botId => this.botService.mountBot(botId))
```
laurentlp, line 140: Why remove the concurrency?

packages/studio-be/src/core/migration/bot-migration-service.ts
```diff
@@ -68,58 +58,66 @@ export class BotMigrationService {
...
+      }
+    })
 
-    logger.warn(chalk`
-${_.repeat(' ', 9)}========================================
-{bold ${centerText(`Migration${migrations.length === 1 ? '' : 's'} Required`, 40, 9)}}
-{dim ${centerText(`Version ${configVersion} => ${this.migService.targetVersion} `, 40, 9)}}
-{dim ${centerText(`${migrations.length} change${migrations.length === 1 ? '' : 's'}`, 40, 9)}}
-${_.repeat(' ', 9)}========================================`)
+    return missing.filter(x => x !== undefined) as MigrationFile[]
```
laurentlp, line 76: PP: Instead of filtering undefined values from the list of missing migrations, you could simply use a for loop and an array of MigrationFile.

packages/studio-be/src/core/migration/bot-migration-service.ts
```diff
@@ -68,58 +58,66 @@ export class BotMigrationService {
...
+    const opts = await this.migService.getMigrationOpts({ botId, botConfig, isDryRun: true })
+
+    const missing = await Promise.mapSeries(migrations, async migration => {
+      try {
+        const result = await this.migService.loadedMigrations[migration.filename].up(opts)
+        if (result.hasChanges) {
+          return migration
+        }
+      } catch (err) {
+        return migration
```
laurentlp, line 72: So if the migration fails, we want to re-run it?

packages/studio-be/src/core/migration/bot-migration-service.ts
```diff
@@ -68,58 +58,66 @@ export class BotMigrationService {
...
+      const label = `${types[info.type]} - ${info.description}`
+
+      try {
+        const result = await this.migService.loadedMigrations[filename].up(opts)
+        debug.forBot(botId, 'Migration step finished', { filename, result })
+
+        if (result.success) {
+          logger.warn(`${chalk.green('[success]')} ${label}`)
+        } else if (result.hasChanges) {
+          logger.error(`${chalk.red('[failure]')} ${label}: ${result.message || ''}`)
```
laurentlp, line 105: `hasChanges` means `failure`?

packages/studio-be/src/core/migration/bot-migration-service.ts
```diff
@@ -68,58 +58,66 @@ export class BotMigrationService {
...
+      hasFailures = true
     })
 
+    logger.warn('')
     if (hasFailures) {
-      return this.logger.error('Could not complete bot migration. It may behave unexpectedly.')
+      return this.logger.error(`[${botId}] Could not complete bot migration. It may behave unexpectedly.`)
     }
 
     await this.configProvider.mergeBotConfig(botId, { version: this.migService.targetVersion })
```
laurentlp, line 121: Here we only update the bot version if a migration failed. Otherwise, we return on line 108.

packages/studio-be/src/core/migration/migration-service.ts
```diff
@@ -159,18 +113,30 @@ export interface MigrationOpts {
   configProvider: ConfigProvider
   database: Database
   inversify: Container
-  metadata: sdk.MigrationMetadata
+  metadata: MigrationMetadata
 }
 
 export type MigrationType = 'database' | 'config' | 'content'
```
laurentlp, line 119: Difference between database, config and content?

packages/studio-be/src/migrations/v12_7_1-1583507028-prune_and_compress_models.ts
```diff
@@ -99,48 +99,44 @@ const migration: Migration = {
...
+    await Promise.mapSeries(botConfig.languages, async lang => {
+      await pruneModels(ghost, lang)
+      const modNames = await listModelsForLang(ghost, lang)
+
+      return Promise.map(modNames, async mod => {
+        try {
+          hasChanges = true
+
+          if (!isDryRun) {
             const model: any = await ghost.readFileAsObject(MODELS_DIR, mod)
```
laurentlp, line 120: The read should also be done in a dry run.

packages/studio-be/src/core/app/botpress.ts
allardy: Because it starts the migration of 5 bots at the same time and every logs gets all messed up, that's the reason why studio migration logs don't make any sense right now. It makes more sense to mount them one after the other. Eventually the goal is to have the studio to work on one bot at a time

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: I'm not so sure about that, technically at this point we're at the dry run, i'm not sure how it should be handled, do we just exit the whole process ? just skip the migration? keep the bot unmounted?

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: If the request was not successful AND there were changes that had to be done, then yes it's a failure

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: Nop, the return of line 108 exits the mapSeries, that was a weird way to avoid adding hasFailures twice, but yeah it's not so clear, i've changed it

packages/studio-be/src/core/migration/migration-service.ts
allardy: Absolutely none, it's just to display the type of change. Also forces us to split our changes in different migrations if we touch separate parts

packages/studio-be/src/migrations/v12_7_1-1583507028-prune_and_compress_models.ts
allardy: It won't change anything but i'll move it to be more concise

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: I'm not sure to get your suggestion. Basically that loop tests all migrations and if it would make changes, then it adds it to the array. Most of the time that array will be filled with undefined values, I prefer to clear out the values before returning the array

packages/studio-be/src/core/migration/bot-migration-service.ts
laurentlp: Alright, and if there were no changes but some errors, we are okay with it? I think it would be a lot simpler if we only had two kind of results: success or failure.

packages/studio-be/src/core/migration/bot-migration-service.ts
```diff
@@ -68,58 +58,66 @@ export class BotMigrationService {
...
+
+        return
+      } catch (err) {
+        logger.attachError(err).error(`${chalk.red('[failure]')} ${label}`)
       }
+
+      hasFailures = true
     })
 
+    logger.warn('')
```
laurentlp, line 114: Is this voluntary?

packages/studio-be/src/core/migration/migration-service.ts
laurentlp: I understand that I mean when would you put database instead of config or content and same for config vs content?

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: if there were no changes then there were no errors, actually i'm not even sure we have migrations returning "success: false" (they would fall into the exception handler)

packages/studio-be/src/core/migration/migration-service.ts
allardy: On the studio it should never be "database", but I left it there because it's the same "system" as the main repo

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: Yep, it adds a new line between migrations, if you had like 10 bots it would have been all cramped up

packages/studio-be/src/core/migration/bot-migration-service.ts
allardy: Changed the behavior. It will display the exception / details and skip that migration. Preventing the mount or exiting the process would be a bit too cumbersome to handle.

And basically, we'll catch those errors in development

packages/studio-be/src/core/migration/bot-migration-service.ts
laurentlp: Maybe add a comment so we don't remove it thinking that it was a mistake?

allardy: I've changed the logic

packages/studio-be/src/migrations/v12_7_1-1583507028-prune_and_compress_models.ts
```diff
@@ -99,48 +99,44 @@ const migration: Migration = {
...
+    botService,
+    ghostService,
+    metadata: { botId, botConfig, isDryRun }
+  }: MigrationOpts): Promise<sdk.MigrationResult> => {
+    let hasChanges = false
+
+    const ghost = ghostService.forBot(botId)
+
+    await Promise.mapSeries(botConfig.languages, async lang => {
+      await pruneModels(ghost, lang)
```
laurentlp, line 112: We should not prune anything when running in dry mode

packages/studio-be/src/core/migration/bot-migration-service.ts
laurentlp: I mean instead of iterating twice on the migration list, you could do it only once. I personally dislike looping over an array to then remove unwanted data afterward. Makes more sense to do it in one shot.

laurentlp:
Also, make sure to run `yarn`, I see changes in the yarn.lock when installing dependencies on this branch!

packages/studio-be/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "start": "cd ./out && cross-env NODE_PATH=./ cross-env BP_MODULES_PATH=B:/bp18/modules node index.js",
-    "build": "yarn clean && yarn tsc",
```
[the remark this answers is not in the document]
allardy: It was useful when the structure changed at the beginning, but now it's more a pain in the ass since it will always fail if you already have a botpress instance running with that studio path (then you have to kill the server, run the command again, then start the server

packages/studio-be/src/migrations/v12_5_0-1577776334-text-markdown.ts
```diff
@@ -24,44 +24,37 @@ const migration: Migration = {
...
+    const entFiles = await bpfs.directoryListing(ELEMENTS_DIR, '*.json')
 
-      for (const fileName of entFiles) {
-        try {
-          const contentElements = await bpfs.readFileAsObject<sdk.ContentElement[]>(ELEMENTS_DIR, fileName)
-          contentElements.forEach(element => updateFormData(element))
+    for (const fileName of entFiles) {
+      try {
+        const contentElements = await bpfs.readFileAsObject<sdk.ContentElement[]>(ELEMENTS_DIR, fileName, {
+          noCache: true
```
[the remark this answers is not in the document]
allardy: When we read the file without altering anything, then we don't need the noCache. The problem is when we mutate the content, if we keep it in cache, then the next run (not the dry one) will not see that theres a change to make

packages/studio-be/src/migrations/v12_5_0-1577776334-text-markdown.ts
```diff
@@ -24,44 +24,37 @@ const migration: Migration = {
...
+        // Just double-checking before writing the content back
+        if (checkFile(fileContent, contentElements.length)) {
+          if (!isDryRun) {
             await bpfs.upsertFile(ELEMENTS_DIR, fileName, fileContent, { ignoreLock: true })
           }
-        } catch (err) {
-          console.error(`Error processing file ${fileName} for bot ${botId}`)
         }
+      } catch (err) {
+        console.error(`Error processing file ${fileName} for bot ${botId}`)
```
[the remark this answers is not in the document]
allardy: Maybe just one file is invalid, so that's why it just outputs a log. I'm not sure we want to skip everything if one file doesn't work

packages/studio-be/src/migrations/v12_5_0-1577776334-text-markdown.ts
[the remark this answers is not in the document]
laurentlp: I think we should never mutate the content when in dry mode. Wouldn't that solve the issue instead of having to be aware of the noCache property and its use?

packages/studio-be/src/migrations/v12_5_0-1577776334-text-markdown.ts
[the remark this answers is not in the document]
laurentlp: When we run database migration, we usually stop the migration if there is an issue with a certain query. I think we should keep our migration logic as consistent as possible.

allardy:
> I think that a better mechanism for the migrations would be to return hasChange = true/false when isDry is true. Otherwise, it should only return success true/false. It makes more sense to me.
> 
> Considering that hasChanges is meant to discover migrations that should be run, it is redundant to return this value again when we run the migrations without the dry run flag.

Good point. Previously we used hasChanges to display a custom message, but it was useless and didnt bring any value, i've adjusted it

packages/studio-be/src/migrations/v12_5_0-1577776334-text-markdown.ts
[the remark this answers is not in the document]
laurentlp: Note: We decided that it's a case-by-case situation and it will be the migration's responsibility to either print an error or throw it.

packages/studio-be/src/migrations/v12_10_0-1591119240-migrate_content_in_flow_data_for_ndu_bots.ts
```diff
@@ -63,52 +63,47 @@ const migration: Migration = {
...
 
-    if (metadata.botId) {
-      await updateFlowContent(metadata.botId)
-    } else {
-      const bots = await botService.getBots()
-      for (const botId of Array.from(bots.keys())) {
-        await updateFlowContent(botId)
+        hasChanges = true
+      } catch (err) {
+        logger
```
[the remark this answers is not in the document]
laurentlp: Note: it's up to the migration to decide if it should fail or not.

PR #202, opened 2021-11-11
Head branch of Bugpress

EFF:
Missing a PR, opening
fix(studio): Node Action overwrite flushing data (#174)


laurentlp:
@michaelmass Not sure what's going on but it doesn't seem to work well on my end. I mean the override feature looks fine, but it seems to break the SmartInput. IIRC I tried to make it work in the past but had a similar issue that I couldn't figure out. the SmartInput component would really need a revamp IMHO.



https://user-images.githubusercontent.com/9640576/141513085-1b4f75a7-4c34-4ae8-b19b-5da8feccda1e.mp4

laurentlp (changes requested):
See my comment above

EFF:
Closing, this. not working at all

PR #208, opened 2021-11-24
chore(dx): better build experience

allardy:
The development experience was not nice with the studio and gulp tasks were a spaghetti mess and it was hard to know how to start. I played around with yarn v2 and it was actually working pretty well, so I decided to give it a try. There are no longer `yarn cmd something` commands to run specific stuff. 

Typing `yarn` will resolve dependencies for all packages; typing `yarn build` will actually call the build command in each package, and typing `yarn watch` will start each package's watcher, and they should build correctly based on their dependency graph.

Don't spent time reviewing the "new files" in `scripts/` folder, those are the exact same commands which were in the gulpfile and I just moved them out of the file. Files in the `.yarn` folder are there to provide a seamless experience for v2 (if you have yarn v1, typing "yarn" will just use the minified cli and v2 will be working)

I'd need other persons to give it a try to be sure it work as expected.

image


packages/studio-be/package.json
```diff
@@ -68,18 +68,14 @@
     "yargs": "^17.0.1",
     "yn": "^2.0.0"
   },
-  "resolutions": {
-    "fstream": ">=1.0.12",
-    "lodash": ">=4.17.19"
-  },
```
laurentlp, line 74: Why remove those resolutions?

packages/ui-shared/package.json
```diff
@@ -52,6 +54,7 @@
     "js-cookie": "^2.2.1",
     "moment": "^2.29.1",
     "react-command-palette": "^0.12.0-0",
-    "react-router": "4.3.1"
+    "react-router": "4.3.1",
+    "studio-be": "*"
```
laurentlp, line 58: Maybe we should think of always naming packages @botpress/<package_name>. So that we are consistent across all repos? Also, this helps differentiate NPM dependencies from packages.

scripts/package.js
```diff
@@ -0,0 +1,21 @@
+require('bluebird-global')
```
laurentlp, line 1: PP: Convert this into a Typescript file and call it using ts-node.

scripts/package.js
```diff
@@ -0,0 +1,21 @@
...
+const path = require('path')
+const fse = require('fs-extra')
+const { execute } = require('./utils/exec')
+
+const package = async () => {
+  const version = require(path.join(__dirname, '../package.json')).version.replace(/\./g, '_')
+
+  try {
+    await execute('yarn', { cwd: './packages/studio-be' })
+    await execute('cross-env pkg package.json', undefined, { silent: true })
```
laurentlp, line 11: Just out of curiosity, is there any reason why cross-env is used here?

laurentlp (commented):
I have some small comments but otherwise, I tested it, and looks fine to me!

packages/studio-be/package.json
allardy: Yarn2 use the resolutions on the main package.json, so these are redundant

scripts/package.js
allardy: I plan to do this, i'm just looking for the right setup so we can use the same in the main repository. I'll open another PR when I get this right

scripts/package.js
allardy: Makes the command run on windows and on linux/mac without having to customize paths.
Eventually i'd like to simplify the packaging process even more, but each steps in its time

laurentlp (approved):

laurentlp (approved):

PR #209, opened 2021-11-25
chore(dev): update typescript

allardy:
Updating TS to the same version as the main repo, and fixing catch unknown warnings


laurentlp (approved):
Tested and LGTM!

PR #212, opened 2021-11-30
feat(cloud): support cloud-ready options in bot config

rob-gordon:
https://linear.app/botpress/issue/DEV-2026/a-user-can-enter-cloud-ready-options-when-creating-a-bot-on-the-admin

Support writing `bot.config.json` with cloud-ready options

### Related PRs

- https://github.com/botpress/typings/pull/2
- https://github.com/botpress/botpress/pull/5704

### DIFF
--- a/packages/studio-be/src/core/bots/bot-service.ts
+++ b/packages/studio-be/src/core/bots/bot-service.ts
@@ -63,7 +63,9 @@ const BotCreationSchema = Joi.object().keys({
       id: Joi.string()
     }
   },
-  locked: Joi.bool()
+  locked: Joi.bool(),
+  isCloudBot: Joi.bool().optional(),
+  bpCloudApiKey: Joi.string().optional()
 })
 
 const debug = DEBUG('services:bots')
--- a/packages/studio-be/src/core/bots/bot-service.ts
+++ b/packages/studio-be/src/core/bots/bot-service.ts
@@ -67,8 +67,8 @@ const BotCreationSchema = Joi.object().keys({
   isCloudBot: Joi.bool().optional(),
   cloud: Joi.object({
     oauthUrl: Joi.string().uri(),
-    clientId: Joi.string().guid(),
-    clientSecret: Joi.string().min(1)
+    clientId: Joi.string().allow(''),
+    clientSecret: Joi.string().allow('')
   }).optional()
 })
 
--- a/packages/studio-be/src/core/bots/bot-service.ts
+++ b/packages/studio-be/src/core/bots/bot-service.ts
@@ -65,7 +65,11 @@ const BotCreationSchema = Joi.object().keys({
   },
   locked: Joi.bool(),
   isCloudBot: Joi.bool().optional(),
-  bpCloudApiKey: Joi.string().optional()
+  cloud: Joi.object({
+    oauthUrl: Joi.string().optional(),
+    clientId: Joi.string().optional(),
+    clientSecret: Joi.string().optional()
+  }).optional()
 })
 
 const debug = DEBUG('services:bots')
--- a/packages/studio-be/src/core/bots/bot-service.ts
+++ b/packages/studio-be/src/core/bots/bot-service.ts
@@ -66,9 +66,9 @@ const BotCreationSchema = Joi.object().keys({
   locked: Joi.bool(),
   isCloudBot: Joi.bool().optional(),
   cloud: Joi.object({
-    oauthUrl: Joi.string().optional(),
-    clientId: Joi.string().optional(),
-    clientSecret: Joi.string().optional()
+    oauthUrl: Joi.string().uri(),
+    clientId: Joi.string().guid(),
+    clientSecret: Joi.string().guid()
   }).optional()
 })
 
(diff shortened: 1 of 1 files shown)


packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -63,7 +63,9 @@ const BotCreationSchema = Joi.object().keys({
       id: Joi.string()
     }
   },
-  locked: Joi.bool()
+  locked: Joi.bool(),
+  isCloudBot: Joi.bool().optional(),
+  bpCloudApiKey: Joi.string().optional()
```
franklevasseur, line 68: yo @rob-gordon 

what are those fields? I've `ctrl+shift+f` in the studio codebase (on branch`next`) and I can't seem to find any usage... 

Is this PR a WIP ? If so, please make it a draft, so we don't review it too soon

rob-gordon: @franklevasseur Apologies! I had this open but hadn't added a reviewer thinking no one would look at it. I'll work with drafts from now on 👍🏻 

These fields are needed for https://github.com/botpress/botpress/pull/5704

I pushed one more commit because the [ticket scope changed](https://linear.app/botpress/issue/DEV-2026#comment-d8c75834) this morning, but the PR is really ready for review now. 

Thanks!

franklevasseur: Alright, no prob!

It's not a thing yet, but since the codebase is all split up, we should start linking PR's of other repo in the description, like:

"This PR goes in hand with this one on the main repo ..."

packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -63,7 +63,13 @@ const BotCreationSchema = Joi.object().keys({
       id: Joi.string()
     }
   },
-  locked: Joi.bool()
+  locked: Joi.bool(),
+  isCloudBot: Joi.bool().optional(),
+  cloud: Joi.object({
+    oauthUrl: Joi.string().optional(),
+    clientId: Joi.string().optional(),
```
laurentlp, line 70: A client Id has a fixed size (UUID), so maybe we should improve the validation we make here. What do you think?

packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -63,7 +63,13 @@ const BotCreationSchema = Joi.object().keys({
       id: Joi.string()
     }
   },
-  locked: Joi.bool()
+  locked: Joi.bool(),
+  isCloudBot: Joi.bool().optional(),
+  cloud: Joi.object({
+    oauthUrl: Joi.string().optional(),
```
laurentlp, line 69: Check that it's a valid URL?

packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -63,7 +63,13 @@ const BotCreationSchema = Joi.object().keys({
...
+    clientId: Joi.string().optional(),
+    clientSecret: Joi.string().optional()
```
laurentlp, line 71: Should have a minimal length IMHO

packages/studio-be/src/core/bots/bot-service.ts
franklevasseur: Can we do that with joi? If not, it'll require some more checks after the joi validation. 

Isn't it slightly overkill? Do you agree?

rob-gordon: @laurentlp url validation here may be overkill given that it's not the user giving us the url. It will be stored in our code base.

https://github.com/botpress/botpress/blob/70bee0c55d2a1d015c51bb57671d977f77bd30c2/packages/bp/src/admin/ui/src/workspace/bots/CreateBotModal.tsx#L132-L136

packages/studio-be/src/core/bots/bot-service.ts
rob-gordon: @laurentlp Updated! 👍🏻

packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -63,7 +63,13 @@ const BotCreationSchema = Joi.object().keys({
       id: Joi.string()
     }
   },
-  locked: Joi.bool()
+  locked: Joi.bool(),
+  isCloudBot: Joi.bool().optional(),
+  cloud: Joi.object({
+    oauthUrl: Joi.string().uri(),
+    clientId: Joi.string().guid(),
+    clientSecret: Joi.string().guid()
```
laurentlp, line 71: Don't think the client secret is a GUID (that would be weird). Maybe you could simply check that it's a certain length. Something like 10 chars. I don't know if that adds any value but knowing the exact size of the key here we could let the user know pretty quickly when its key is definitely invalid. What do you think?

rob-gordon: The client secret and client id are both UUIDv4, and `guid()` is the [joi alias](https://joi.dev/api/?v=17.5.0#stringguid---aliases-uuid) for validating them. I got all this info from @michaelmass in case you want to confirm 👍🏻

laurentlp (approved):

PR #214, opened 2021-12-01
chore(dx): normalize package names & use ts scripts

allardy:
Few DX changes:

- Normalize package names: @botpress/package
- Convert scripts to TS & use ts-node instead
- Small cleanup


packages/ui-shared/package.json
```diff
@@ -50,11 +50,11 @@
     "dist"
   ],
   "dependencies": {
+    "@botpress/studio-be": "*",
```
laurentlp, line 53: Why does ui-shared depend on the studio's backend?

scripts/package.ts
```diff
@@ -1,9 +1,10 @@
-require('bluebird-global')
-const path = require('path')
-const fse = require('fs-extra')
-const { execute } = require('./utils/exec')
+import 'bluebird-global'
```
laurentlp, line 1: Is bluebird still required?

scripts/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "strict": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "noImplicitAny": false,
+    "experimentalDecorators": true,
```
laurentlp, line 8: Is this required? If so, by what?

scripts/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "target": "es2019",
+    "module": "commonjs",
```
laurentlp, line 10: I think you can remove this property or at least you could use a more recent type of module (e.g. es6).

scripts/package.ts
```diff
@@ -14,8 +15,11 @@ const package = async () => {
...
-    console.error('Error running: ', err.cmd, '\nMessage: ', err.stderr, err)
+    if (err instanceof Error) {
+      const error = err as ExecException
+      console.error('Error running: ', error.cmd, '\nMessage: ', error['stderr'], err)
+    }
   }
 }
 
-package()
+void packageApp()
```
laurentlp, line 25: Not sure I understand why you added the void here?

laurentlp (commented):
Small comment but overall looks good to me!!! Awesome work here!!

packages/ui-shared/package.json
allardy: Because it uses common files which are generated in the backend. Eventually i'd like to move those common files in a "common" package (once the typings submodule is added, because theres a couple of sdk typings from the sdk)

That dependency was always there, but we controlled it with the gulp parallel/series, it's much better to have a straightforward dependency

scripts/package.ts
allardy: For Promise.forCallback

scripts/tsconfig.json
allardy: I've copy-pasted from elsewhere, but sure I can remove it

scripts/package.ts
allardy: Top-level awaits are not permitted, i used void instead of tslint ignore

scripts/tsconfig.json
allardy: It might be the default, but we have it defined in all our tsconfig files. Also, es6 is kinda old (2015), we currently use es2019

laurentlp (dismissed):
There seems to have issues with installing dependencies in the code style actions, but else LGTM

PR #220, opened 2021-12-07
chore(dx): run prerelease task only when pr is ready for review

franklevasseur:
Alright, this is my take on improving the prerelease GH task.

If you toggle between "draft" and "ready for review" you'll see that the action gets canceled and started accordingly.

Also, you'll notice that `${GITHUB_HEAD_REF}` outputs the correct branch name instead of something like `refs_pull_120_merge`.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Extract Informations
         id: info
-        uses: botpress/gh-actions/extract_info@master
+        uses: botpress/gh-actions/extract_info@v1
         with:
           branch: ${{ steps.branch.outputs.branch_name }}
 
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -30,7 +30,7 @@ jobs:
         id: info
         uses: botpress/gh-actions/extract_info@master
         with:
-          branch: ${{ github.ref_head }}
+          branch: ${{ env.GITHUB_HEAD_REF }}
 
       - name: Debug
         run: |
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Debug
         run: |
           echo if this action succeeds a binary will be available at
-          echo head branch is ${{ env.GITHUB_HEAD_REF }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
```diff
@@ -55,4 +72,4 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         run: |
-          aws s3 sync bin s3://botpress-dev-bins/studio/${{ steps.info.outputs.branch_sanitized }}
+          aws s3 sync bin s3://botpress-dev-bins/studio/${GITHUB_HEAD_REF}
```
allardy, line 75: Can we pass this reference to extract_info instead? The reason we use "branch_sanitized" is because we can't use the branch name directly, branches with a slash cause problem. This work for simple branches, but there's branches like "features/DEV-1240". That action converts any non-word with an underscore, so we can safely handle it, and the downloader also uses the same logic

.github/workflows/pre-release.yml
```diff
@@ -43,10 +64,6 @@ jobs:
       - name: Package Binaries
         run: yarn package
 
-      - name: Extract Informations
-        id: info
-        uses: botpress/gh-actions/extract_info@master
-
       - name: Rename Binary Files
         uses: botpress/gh-actions/rename_binaries@master
```
franklevasseur, line 68: I also forgot that `botpress/gh-actions/rename_binaries` renames binary with incorrect name...

.github/workflows/pre-release.yml
```diff
@@ -26,10 +26,17 @@ jobs:
     name: Build and Publish Binaries
     runs-on: ubuntu-latest
     steps:
+      - name: Extract Informations
+        id: info
+        uses: botpress/gh-actions/extract_info@master
```
laurentlp, line 31: Please use v1 instead of master

franklevasseur: criss

franklevasseur:
this works

franklevasseur:
I merge:

- this works
- this was broken before
- its only next, we can still change it
- if you dont agree DM me on Slack

PR #222, opened 2021-12-09
feat(nlu): use cloud nlu when cloud config is set

franklevasseur:
title says it all

goes in hand with: https://github.com/botpress/botpress/pull/5750

closes DEV-2027


EFF (approved):
Yess !

PR #223, opened 2021-12-10
fix(nlu): trim entity occurences on change

franklevasseur:
fixes DEV-1217
fixes https://github.com/botpress/botpress/issues/4446

entities_trim

### DIFF
--- a/packages/studio-ui/src/web/views/Nlu/entities/ListEntityOccurrence.tsx
+++ b/packages/studio-ui/src/web/views/Nlu/entities/ListEntityOccurrence.tsx
@@ -24,10 +24,7 @@ export const Occurrence: FC<Props> = props => {
       }
       placeholder={lang.tr('nlu.entities.synonymPlaceholder')}
       rightElement={<Button icon="delete" minimal onClick={props.remove} />}
-      onChange={(synonyms: string[]) => {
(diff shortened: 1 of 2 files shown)


packages/studio-ui/src/web/views/Nlu/entities/ListEntityOccurrence.tsx
```diff
@@ -24,7 +24,10 @@ export const Occurrence: FC<Props> = props => {
       }
       placeholder={lang.tr('nlu.entities.synonymPlaceholder')}
       rightElement={<Button icon="delete" minimal onClick={props.remove} />}
-      onChange={(synonyms: string[]) => props.onChange({ ...props.occurrence, synonyms })}
+      onChange={(synonyms: string[]) => {
+        synonyms = synonyms.map(s => s.trim())
```
rob-gordon, line 28: Why do we need to re-trim them on every change? Is this in order to fix older occurrences that weren't trimmed?

rob-gordon (dismissed):
It may be easier to compare against trimmed strings in `isUniqueInEntity` and then update untrimmed strings once on save, rather then on each onChange event - but N/B

Looks good!

rob-gordon (approved):

PR #224, opened 2021-12-13
Remove enterprise callout for enterprise user

rob-gordon:
Don't show upsell to users already running enterprise

Screen Shot 2021-12-13 at 11 35 42 AM

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
+++ b/packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
@@ -27,11 +27,15 @@ const Toolbar: FC<Props> = props => {
   return (
     <header className={style.toolbar}>
       <div className={style.list}>
-        <Tooltip position="right-bottom" content={lang.tr('studio.flow.toolbar.salesCallToActionDescription')}>
-          <a className={style.cta_btn} target="_blank" href="https://botpress.com/free-trial?ref=bp-studio">
-            {lang.tr('studio.flow.toolbar.salesCallToAction')}
-          </a>
-        </Tooltip>
+        {window.IS_PRO_ENABLED ? (
+          <span />
+        ) : (
+          <Tooltip position="right-bottom" content={lang.tr('studio.flow.toolbar.salesCallToActionDescription')}>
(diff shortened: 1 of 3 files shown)


packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
```diff
@@ -27,11 +27,15 @@ const Toolbar: FC<Props> = props => {
...
-        </Tooltip>
+        {window.IS_PRO_ENABLED ? (
+          <span />
+        ) : (
+          <Tooltip position="right-bottom" content={lang.tr('studio.flow.toolbar.salesCallToActionDescription')}>
+            <a className={style.cta_btn} target="_blank" href="https://botpress.com/request-trial-from-app">
+              {lang.tr('studio.flow.toolbar.salesCallToAction')}
+            </a>
+          </Tooltip>
+        )}
```
laurentlp, line 38:
```suggestion
        {!window.IS_PRO_ENABLED && (
          <Tooltip position="right-bottom" content={lang.tr('studio.flow.toolbar.salesCallToActionDescription')}>
            <a className={style.cta_btn} target="_blank" href="https://botpress.com/request-trial-from-app">
              {lang.tr('studio.flow.toolbar.salesCallToAction')}
            </a>
          </Tooltip>
        )}
```

laurentlp (commented):
Small suggestion, otherwise, LGTM!

packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
rob-gordon: Because our styles rely on `space-between` to position the emulator and debugger buttons to the right (in juxtaposition to this element on the left), returning no element means we would have to write another set set of styles to accomplish the same thing when the Enterprise button isn't present 👎🏻  

I feel like one un-semantic span outweighs two sets of styles to accomplish the same thing but if you disagree I'm happy to write them!

https://github.com/botpress/studio/blob/72f26ed55cca3b081270207795fbbf186eb960f5/packages/studio-ui/src/web/components/Layout/Toolbar/style.scss#L21-L24

laurentlp (approved):

PR #225, opened 2021-12-14
fix(copy): copying skill makes clone but uses same cms content

rob-gordon:
#### Before
Copying a skill node does a shallow copy, creates a new node linked to same skill

#### After 
Copying a skill node creates a clone of the skill, creates a new node linked to clone, but keeps shallow reference to CMS content in skill

---

This behavior was determined to be the most predictable for current users.

### DIFF
--- a/packages/studio-ui/src/web/reducers/flows.ts
+++ b/packages/studio-ui/src/web/reducers/flows.ts
@@ -38,6 +38,7 @@ import {
   updateFlowProblems
 } from '~/actions'
 import { hashCode, prettyId } from '~/util'
+import { copyName } from '~/util/flows'
 
 export interface FlowReducer {
   currentFlow?: string
@@ -180,23 +181,6 @@ const popHistory = stackName => state => {
   return {
     ...newState,
     currentHashes
-  }
-}
-
-const copyName = (siblingNames, nameToCopy) => {
-  const copies = siblingNames.filter(name => name.startsWith(`${nameToCopy}-copy`))
-
-  if (!copies.length) {
-    return `${nameToCopy}-copy`
-  }
-
-  let i = 1
-  while (true) {
-    if (!copies.find(name => name === `${nameToCopy}-copy-${i}`)) {
-      return `${nameToCopy}-copy-${i}`
-    } else {
-      i += 1
-    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/studio-ui/src/web/reducers/flows.ts
```diff
@@ -534,7 +518,7 @@ reducer = reduceReducers(
           id: `skill-${flowRandomId}`,
           type: 'skill-call',
           skill: skillId,
-          name: `${skillId}-${flowRandomId}`,
+          name: payload.nodeName ?? `${skillId}-${flowRandomId}`,
```
laurentlp, line 521: How could and would `nodeName` be undefined here? Also, what is the outcome if it is? The skill node would have no name?

laurentlp (commented):
LGTM Nice work!!!!

The only small "problem" I've is that if we update a content-element, it only refreshes the skill where we update the element, not the others.


https://user-images.githubusercontent.com/9640576/148804697-8bcbfc42-06ac-4a21-b228-b453087b7831.mp4

packages/studio-ui/src/web/reducers/flows.ts
rob-gordon: I had to investigate this because I forgot everything over the holidays 😆 

The `requestInsertNewSkill` action is called in two different thunks:

When we create a new skill and insert it, this is called:
https://github.com/botpress/studio/blob/e5bd786a809ffe3cf9c4927022a7be7a3871a975/packages/studio-ui/src/web/views/FlowBuilder/skills/index.jsx#L79-L85
https://github.com/botpress/studio/blob/b957b5b1f505a1443b9e8b30bf018891dbab1339/packages/studio-ui/src/web/actions/index.ts#L398-L401

and when we copy an existing skill and paste it, this is called:

https://github.com/botpress/studio/blob/b957b5b1f505a1443b9e8b30bf018891dbab1339/packages/studio-ui/src/web/actions/index.ts#L212-L219

Creating a new skill doesn't contain a nodeName, so we generate one. Pasting a skill needs to generate the "${sourceNodeName} Copy 1" nodeName, so we pass one.

laurentlp (approved):

PR #229, opened 2021-12-21
fix(studio-be): fix cache invalidation being discarded by flow service

laurentlp:
This PR should fix an issue where overriding a bot with the same id (or uploading a bot with an id that already existed) would result in having outdated flow content.

Here is the reason why we should check for matches with and without the prefix `/data`: [botpress/studio/#2](https://github.com/botpress/studio/pull/2/files#diff-f6d6bd05cc1c54dd9eab8e84d2cf52606379fd19873f3ae2ccf344023def0407R41)

This endpoint (`/invalidateFile`) is called whenever there is an invalidation in Botpress. Since we remove the `data/` part from the key when running using SQLite, the flow-service would discard the cache invalidation resulting in outdated flow data being used.

I also updated the config loader and hint service to make sure that they watch for both keys containing or not the `data/` part.

Closes https://github.com/botpress/botpress/issues/5769
Closes DEV-2203

## Testing

1. To reproduce the issue, create an `on_stage_request` hook using this code:

```js
  const STAGE_SEPERATOR = '__'
  const VERSION_SEPERATOR = ' :: v'
  const botId = bot.id.split(STAGE_SEPERATOR)[0]
  const STAGE_KEY = `${botId}-PipelineVersion`
  const botLogger = bp.logger.forBot(bot.id)

  const getPromoteVersionStage = () => {
    if (pipeline.length < 2) return 'NEVER'
    let stage2 = pipeline[1]
    return stage2.id
  }

  const getNextVersion = async () => {
    const promoteVersion = getPromoteVersionStage() === bot.pipeline_status.stage_request.id

    let globalKvs = bp.kvs.global()
    let storeExists = await globalKvs.exists(STAGE_KEY)
    let curVersion = storeExists ? await globalKvs.get(STAGE_KEY) : 0

    if (!promoteVersion) return curVersion

    let nextVersion = curVersion + 1

    await globalKvs.set(STAGE_KEY, nextVersion)

    return nextVersion
  }

  const getStageId = () => [botId, bot.pipeline_status.stage_request.id].join(STAGE_SEPERATOR)

  const getStageName = async () => {
    let botName = bot.name.split(VERSION_SEPERATOR)[0]
    let nextVersion = await getNextVersion()

    return `${botName}${VERSION_SEPERATOR}${nextVersion}`
  }

  const handler = async () => {
    try {
      let stageId = getStageId()
      let stageName = await getStageName()

      bot.id = stageId
      bot.name = stageName
    } catch (err) {
      botLogger.error(JSON.stringify(err, null, 2))
    }
  }

  return handler()
```

2. Make sure you add at least one other stage to your Bot Pipeline.
3. Create an empty bot and add two nodes with content to it.
4. Promote your bot to the next stage.
5. Both bots should work properly.
6. Remove one of the nodes in the original bot.
7. Promote it again
8. This time the promoted bot will still display the deleted node.

You can then repeat these steps to see that the issue is not resolved.


laurentlp:
@allardy I need more details on why we have to handle this when we call `/invalidateFile` :

```typescript
if (process.BPFS_STORAGE === 'disk') {
  key = key.replace('::data/', '::')
}
```

We might want to move this logic inside a util function so that all services listening to invalidation can select the files they want to handle accordingly

packages/studio-be/src/core/dialog/flow/flow-service.ts
```diff
@@ -174,6 +174,11 @@ export class ScopedFlowService {
     const expectedSaves = this.expectedSavesCache.get(flowPath)
 
     if (!expectedSaves) {
+      // fix an issue when creating a bot where the .flow.json is written but not the .ui.json because of the locking mechanism
+      if (!(await this.ghost.fileExists(FLOW_DIR, this.toUiPath(flowPath)))) {
+        return
+      }
+
```
laurentlp, line 181: This was added in the ghost of Botpress, but we forgot to copy it in the studio's ghost.

packages/studio-be/src/core/dialog/flow/flow-service.ts
```diff
@@ -93,7 +93,7 @@ export class FlowService {
   private _listenForCacheInvalidation() {
     this.cache.events.on('invalidation', async key => {
       try {
-        const matches = key.match(/^([A-Z0-9-_]+)::data\/bots\/([A-Z0-9-_]+)\/flows\/([\s\S]+(flow)\.json)/i)
+        const matches = key.match(/^([A-Z0-9-_]+)::bots\/([A-Z0-9-_]+)\/flows\/([\s\S]+(flow)\.json)/i)
```
laurentlp, line 96: This is the part that need some clarification. In certain cases, the `data/` part is omitted but I'm not quite sure why.

samuelmasse (approved):
LGTM

PR #230, opened 2021-12-22
feat(superinput): add styles to superinput

rob-gordon:
Here's an overview of the base styles. One thing of note, I tried using the CodeMirror's javascript styling API, but I had weird behavior and ultimate found it more cumbersome than plain old CSS. If we think that will be an issue we can try to port it into the styling API but it worked well for me in development.

I tried to inherit from BlueprintJs as much as possible so that it will be visually seamless with the app.

@Raid55 I did change what the sample app interface looks like, just to have more screen real estate, but I'm happy to change it back if you preferred it the other way!

**Basic Focus State**
Screen Shot 2021-12-22 at 11 10 54 AM

**Valid Preview**
Screen Shot 2021-12-22 at 11 11 11 AM

**Multiline Valid Preview**
Screen Shot 2021-12-22 at 11 11 20 AM

**Invalid Preview**
Screen Shot 2021-12-22 at 11 12 00 AM

**Context Menu / InfoCard**
> Note: I only styled the wrapper of the InfoCard here because the content in the InfoCard felt like a placeholder. As soon as we know what we can to include in the InfoCard I can style it accordingly.

Screen Shot 2021-12-22 at 11 12 21 AM

**InspectCard**
Screen Shot 2021-12-22 at 12 15 16 PM

---

#### Evaluates to...

Screen Shot 2022-01-18 at 9 38 04 AM

#### Docs Link

https://user-images.githubusercontent.com/159949/149962055-6a072c50-9442-4d5b-b158-9de712683c2f.mov

#### Updated CodeMirror syntax highlighting

Sadly CodeMirror only differentiates between parent and property (not like function and arguments).
Screen Shot 2022-01-18 at 9 40 22 AM

## Other Updates
- had to write a small utility for showing the evaluates to string, because we were losing information for booleans that were false for instance.
- got rid of dependency on ui-shared colors because ui-shared may not always be there! :)


EFF (commented):
Awesome ! Kindof about time we start our own ui-kit and ditch ui-shared :)

packages/super-input/src/SuperInput/extensions/autocomplete/index.ts
```diff
@@ -79,7 +79,7 @@ const globsCompletions = (globs: any) => {
         .replace(']', '')
         .split(/\.|\?\./)
 
-      const from = /[\.\[]/.test(nodeBefore.name) ? nodeBefore.to : nodeBefore.from
+      const from = /[.[]/.test(nodeBefore.name) ? nodeBefore.to : nodeBefore.from
```
Raid55, line 82: does this evaluate the dot literally since it's inside `[]`?

Raid55 (approved):

packages/super-input/src/SuperInput/extensions/autocomplete/index.ts
Raid55: can we remove the `\` on the `[` also?

rob-gordon: Yeah, exactly. I didn't catch this with my own eyes or anything, webpack was throwing a warning about it being unnecessary so I just cleaned it up.

PR #232, opened 2022-01-11
fix(layout): bottom panel maintain size

rob-gordon:
Before

https://user-images.githubusercontent.com/159949/149004132-413dd269-f6c3-4c98-bf7a-949406071d77.mov

After

https://user-images.githubusercontent.com/159949/149004146-35e02dcb-a0d7-4ce3-8d8f-3db1446e0331.mov


laurentlp (approved):
Nice work! LGTM!!!

PR #233, opened 2022-01-12
chore(flow): prevent using same flow name

EFF:
This PR goes hand in hand with https://github.com/botpress/botpress/pull/5771

Prevent creating or renaming a flow with an already existing flow name. Case insensitive.


davidvitora (approved):
LGTM

PR #234, opened 2022-01-12
fix(sidebar): remove modules if cloud bot

rob-gordon:
Remove modules from sidebar if working on Cloud Enabled bot

Screen Shot 2022-01-12 at 9 24 12 AM


packages/studio-ui/src/web/components/Layout/Sidebar.tsx
```diff
@@ -133,7 +133,8 @@ const Sidebar: FC<Props> = props => {
 
 const mapStateToProps = (state: RootReducer) => ({
   viewMode: state.ui.viewMode,
-  modules: state.modules
+  modules: state.modules,
+  isCloudBot: Boolean(state.bot?.cloud?.clientId)
```
EFF, line 137: one could create a create a cloud ready bot but not enter client ID just yet. That could would return false even though it's a cloud bot. 

Isn't there a boolean set in the bot config ?

EFF (changes requested):
Not guaranteed to work

laurentlp (approved):
LGTM!

PR #235, opened 2022-01-12
fix(actions): hide module actions on cloud-enabled bots

rob-gordon:
Hide module actions on cloud-enabled bots because they're not supported on the cloud yet

Screen Shot 2022-01-12 at 9 47 47 AM

### DIFF
--- a/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ActionModalForm.tsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ActionModalForm.tsx
@@ -304,7 +304,7 @@ class ActionModalForm extends Component<Props, State> {
 
 const mapStateToProps = (state: RootReducer) => ({
   actions: state.skills.actions?.filter(a => a.legacy),
-  isCloudBot: Boolean(state.bot?.cloud?.clientId)
+  isCloudBot: Boolean(state.bot.isCloudBot)
 })
 
 export default connect(mapStateToProps, undefined)(ActionModalForm)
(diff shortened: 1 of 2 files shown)


packages/studio-ui/src/web/views/FlowBuilder/nodeProps/ActionModalForm.tsx
```diff
@@ -99,13 +99,18 @@ class ActionModalForm extends Component<Props, State> {
...
-        return {
-          label: x.name,
-          value: x.name,
-          metadata: { ...x }
-        }
-      })
+      avActions: (this.props.actions || [])
+        // hide module actions on cloud-enabled bots
+        .filter(action =>
+          !this.props.isCloudBot ? true : action.name.indexOf('builtin') === 0 || action.name.indexOf('/') === -1
```
franklevasseur, line 105: No need for ternary operator:
```ts
!this.props.isCloudBot || action.name.indexOf('builtin') === 0 || action.name.indexOf('/') === -1
```

Also, **personal pref**: I would extract 2 small private functions `isBuiltinAction()` and `isModuleAction()` so your condtion looks like this:
```ts
!this.props.isCloudBot || isBuiltinAction(action) || !isModuleAction(action)
```

franklevasseur (commented):
I haven't tested it, but just reading the code it LGTM. I just have one question:

Are we sure, that, except for the builtin module, **all** module actions and **only** module actions have a forward slash `/` in their name?

rob-gordon:
> Are we sure, that, except for the builtin module, all module actions and only module actions have a forward slash / in their name?

@franklevasseur That logic came from F's comment in the linear task [here](https://linear.app/botpress/issue/DEV-2033#comment-555d920c)

franklevasseur (approved):
In that case, it's all good!

PR #242, opened 2022-01-17
feat(studio): allow disabling file listeners for dumb deployments

davidvitora:
Related to https://github.com/botpress/botpress/pull/5664

Since the Studio is started with Botpress core, even though it's not used in DDs, this becomes a problem.


davidvitora:
@allardy maybe, in this case, is better to just disable the studio, whats is your opinion on this?

samuelmasse:
Closing due to inactivity. Feel free to reopen on new commits

davidvitora:
This fix was tested and it is working

laurentlp:
> Related to [botpress/botpress#5664](https://github.com/botpress/botpress/pull/5664)
> 
> Since the Studio is started with Botpress core, even though it's not used in DDs, this becomes a problem.

DD? Also, how does it becomes a problem and how does your fix solves the issue?

davidvitora:
@laurentlp same as the related PR that was done a while ago for the Core,  basically containers on AWS and other services have a limited amount of watchers that can't be changed, since DDs (Dumb Deployments) store all files on the disk, Botpress breaks. This was fixed in the core but the studio also create watchers for files.

packages/studio-be/src/core/bpfs/cache-invalidators.ts
```diff
@@ -46,25 +47,30 @@ export namespace CacheInvalidators {
     install(objectCache: ObjectCache) {
       this.cache = objectCache
 
+      // Support for DDs
```
laurentlp, line 50:
```suggestion
      // Support for Dumb Deployments (DD)
```

laurentlp (commented):
Small comments otherwise LGTM

laurentlp (approved):

PR #251, opened 2022-01-24
fix(builtin): ensure smalltalk qnas have contexts

franklevasseur:
WIP

fixes: BUS-188

There was 2 bugs:
1. smalltalk QNAs had no contexts, so intents made of these had no contexts either. The NLU Server does not allow that (as it should)
2. NLU Server running on cloud has a better error reporting mechanism than the one running in binary, so errors looked like `[Object object]`... This PR supports both error format.


EFF (approved):
well done

PR #254, opened 2022-01-26
feat(superinput): code-editor-like input box

Raid55:
This is the pull request to merge the superinput package with the rest of the studio (into the next branch).

Screen Shot 2022-01-25 at 6 33 52 PM

This PR does not include implementation into `studio-ui`.


allardy:
The SuperInput doesn't seems to be used anywhere on the studio, could we implement it in at least one place before merging on next? I would love to try this in a real-world situation :)

packages/superinput/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+```typescript
+enum SI_TYPES { // types of superinput
+  TEMPLATE = 0, // for template strings using the {{}} delimiter
+  EXPRESSION = 1, // for full javascript expressions that evaluate its full output
+  BOOL = 2 // for full javascript expressions that evaluate its output to a bool
+}
+
+export interface ISiProps {
+  globs?: any // variables to pull eval info from, must be IO.IncomingEvent
```
allardy, line 78: A glob for me is like `some/path/**`, maybe something like "data" would be more generic ? Also not sure to understand "must be IO.IncomingEvent", if it is supposed to be that type, why use any ?

packages/superinput/README.md
```diff
@@ -0,0 +1,121 @@
...
+## Props
+
+---
+
+```typescript
+enum SI_TYPES { // types of superinput
+  TEMPLATE = 0, // for template strings using the {{}} delimiter
+  EXPRESSION = 1, // for full javascript expressions that evaluate its full output
+  BOOL = 2 // for full javascript expressions that evaluate its output to a bool
+}
```
allardy, line 75: Could we use text intead of numbers? it's easier to use (also, please avoid the prefix "I" for interfaces, we have typings, its not necessary)
```js
type SuperInputType = 'template' | 'expression' | 'bool' 

interface SuperInputProps {
  type: SuperInputType
}
```

packages/superinput/src/Superinput.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { placeholder as placeholderExt, keymap, EditorView, ViewUpdate } from '@codemirror/view'
+import React, { useEffect, useState, useRef, MutableRefObject } from 'react'
+
+import EditorFrame from './EditorFrame'
+import EvalPanel from './EvalPanel'
+import { bpAutocomplete, BPLang, hoverInspect, exprDecorator } from './extensions'
+import { SI_TYPES, ISiProps } from './types'
+import { isError, evalStrTempl } from './utils/tokenEval'
+
+export default function Superinput({
```
allardy, line 16: SuperInput

packages/superinput/src/Superinput.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setView(newView)
+
+    if (autoFocus) {
+      newView.focus()
+    }
+
+    return () => {
+      if (view) {
+        view.destroy()
+      }
```
allardy, line 82: just a detail, with optional chaining you can simply use `view?.destroy()` instead of checking if it is not null before

packages/superinput/README.md
Raid55: I can change the name, I named it globs for global variables. Data could be a little open-ended, could a more specific name be better, like `eventState` or something?

For the second one, "must be IO.IncomingEvent" is half true. You can put anything there and autocomplete will work, however to get the typescript docs and types it there must be a parent or child of `IO.IncomingEvent` since we generate the docs before hand.

EFF (commented):
few questions

samuelmasse:
@Raid55 merged next into. There was a little conflict in a file but appart from that I don't think I changed much

packages/superinput/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import SuperInput from './SuperInput'
+import { SI_TYPES } from './types'
```
samuelmasse, line 2: Should be
```ts
import SuperInput from './Superinput'
```
With a lowercase i

samuelmasse (commented):
I'm not able to run yarn build on the superinput package so far. There's ony tiny error in the code at the index.ts file with an incorrect import. When I fix that rollup is able to create the cjs and esm lib but it finishes with a random error :
Screenshot from 2022-02-04 12-10-18

samuelmasse:
We did some changes with the prettier version and setup, so there are some formatting errors to fix. Check the Codestyle action to see the files. Should be able to fix just by doing CTRL + S in each file

packages/superinput/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "react-dom": "16.8.6",
+    "rollup": "^2.64.0",
+    "rollup-plugin-dts": "^4.1.0",
+    "rollup-plugin-peer-deps-external": "^2.2.4",
+    "rollup-plugin-postcss": "^4.0.2",
+    "rollup-plugin-terser": "^7.0.2",
+    "ts-node": "^10.4.0",
+    "tslib": "^2.3.1",
+    "typescript": "^4.1.2",
+    "typescript-json-schema": "^0.53.0"
```
samuelmasse, line 79: We'll setup typescript as a yarn sdk pretty soon, so we'll remove that eventually but it's not blocking to merge this PR.

packages/superinput/src/types.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface ISiProps {
+  value?: string
+  onChange?: (newValue: string) => any
+  placeholder?: string
+  type?: SI_TYPES
+  globs?: any
+  noGlobsEvalMsg?: string
+  autoFocus?: boolean
+}
```
samuelmasse, line 20: Minor comment about coding conventions. We don't usually prefix our interfaces with `I`. Also enums should be CamelCase

packages/superinput/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "react-json-view": "^1.21.3",
+    "tree-model": "^1.0.7",
+    "web-vitals": "^1.0.1"
+  },
+  "scripts": {
+    "build": "yarn doctree && rollup -c",
+    "dev": "rollup -c -w",
+    "demo": "parcel demo/index.html",
+    "doctree": "ts-node --skip-project ./scripts/genDocTree.ts"
+  },
```
samuelmasse, line 37: Something weird I noticed is that the SuperInput package has two dev dependencies for two different bundlers. It uses parcel for the demo, but it uses rollup to produce the lib. Parcel could also produce a bundle using parcel build. I think it would be best to avoid extra dependencies when possible.

If we consider that the ui-studio is using webpack to bundle itself as well, we have a total of 3 bundlers in this repo. I'm thinking of adjusting the studio so it can use parcel instead. We also use babel in the studio which I'm thinking of removing as well.

So ideally we would be using tsc for the backend, with ts-node for development, and parcel for the frontend to bundle and for development. We should also make sure that the code in the ui packages can be compiled using tsc from time to time, since tsc is often stronger for type checking than bundlers like rollup and parcel. It can also spot issue in the project configuration (like the tsconfig), that bunlders sometime ignore.

@EFF @laurentlp @rob-gordon @Raid55 what do you think.

This is not necessary to resolve to merge this PR, but it's something to thing about for the repo as a whole.

packages/superinput/scripts/genDocTree.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+const docs = {
+  docTree,
+  fallback
+}
+
+fs.writeFileSync(join(__dirname, '../src/docsTree.json'), JSON.stringify(docs))
+// eslint-disable-next-line
+console.log(`Generated doctree at ${__dirname + '../src/docsTree.json'}`)
```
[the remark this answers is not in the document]
Raid55: I added the console log so it can be a little verbose when this step is done, the eslint comment is because we don't allow console.log.

I can remove it if its not needed.

packages/superinput/src/EvalPanel.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+import { IPanelProps } from './types'
+
+const EvalPanel = ({ valid, text }: IPanelProps) => {
+  return (
+    <div
+      className={`bp-editor-panel ${valid === null ? 'no-globs' : valid ? 'valid' : 'invalid'}`}
+      // onMouseDown={e => {}}
```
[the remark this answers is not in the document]
Raid55: I left this here, it was a test, onMouseDown to copy the result onto clipboard. but il remove it we can add that feature later

samuelmasse (approved):
LGTM

PR #256, opened 2022-01-26
feat(layout): added switches to toggle different panels

EFF:
If you look at the file changes, this one looks a bit bigger than it is.

If you look at the final changes, you would be surprised it needed all those changes.

### Done
0- renamed the ToolBar component for TopNav (as we have other Toolbar components all around got me confused)
1- broke down the top nav in smaller components, removed logic
2- keep state of explorer (left panel) in global app state
3- use the same component for Container and Explorer panel so that the state is kept between pages and page refresh
4- reworked translations files a little bit (just for top nav)
5- some logic to render the cmd or ctrl 

### Ouf of scope of this PR
**I almost went into doing them**
1- review the main layout (Layout/index.tsx) file, it's a mess it would have require a lot of refactor and would have bloat the pr
2- change the window listeners for emulator toggling for a the behavior that is exactly like the 2 other panels
3- re(move) the emulator top right button. As they will eventually change for train and deploy buttons
4- get rid of SplitPane library
5- extract shortcuts in a single place
6- replace all occurence of cmd/ctrl by the logic I wrote

closes DEV-2247

Preview:

https://user-images.githubusercontent.com/955524/151258003-de14c361-5086-4cb9-b73d-1cb4f4314cfb.mp4


packages/studio-ui/src/web/components/Layout/TopNav/RightToolbar.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { RootReducer } from '~/reducers'
+
+import style from './style.scss'
+
+interface Props {
+  docHints: any[]
+  isEmulatorOpen: boolean
+}
+
+const RightToolBar: FC<Props> = props => {
```
rob-gordon, line 15: No change necessary but worth mentioning- people generally don't use `FC` anymore. You would just do `const RightToolBar = (props: Props) => ...`
https://github.com/facebook/create-react-app/pull/8177
https://www.harrymt.com/blog/2020/05/20/react-typescript-react-fc.html
https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/

packages/studio-ui/src/web/components/Layout/TopNav/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type Props = ReturnType<typeof mapStateToProps> & typeof mapDispatchToProps
+
+const Toolbar: FC<Props> = props => {
+  // We cannot use setEmulatorOpen as it's actually the state is actually controlled by the message passed from the iframe
+  // Will need some refactor for the emulator
+  // This should be moved this somewhere else, it seems to be all around the app
+  const toggleEmulator = () => {
+    window.botpressWebChat.sendEvent({ type: 'toggle' })
+  }
```
rob-gordon, line 23: This could be moved outside of the component / doesn't need to be created every render

packages/studio-ui/src/web/components/Layout/TopNav/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <nav className={style.topNav}>
+      <EnterPriseTrial />
+      <div className={style.layoutControls}>
+        <ToolTip
+          content={lang.tr('topNav.toggleExplorer', { shortcut: `${utils.shortControlKey} B` })}
+          position={Position.BOTTOM}
+        >
+          <Button
+            onClick={props.toggleExplorer}
```
rob-gordon, line 34: For these buttons you can add the `minimal` prop ([reference](https://blueprintjs.com/docs/#core/components/button)). Then you need a lot less css

```scss
.layoutControls {
    button {
      color: var(--c-text--disabled);

      &:hover,
      &.active {
        color: var(--shark);
      }

      &:hover {
        background: var(--bg);
      }
    }
  }
```

rob-gordon (approved):
Looks good!

PR #257, opened 2022-01-27
fix(ui): no model warning in debugger only appears when there's no model

franklevasseur:
fixes: DEV-219

https://user-images.githubusercontent.com/25970722/151442001-a13eef9f-bf6a-4251-9c74-f546313979a1.mov

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/BottomPanel/Debugger/views/Summary.tsx
+++ b/packages/studio-ui/src/web/components/Layout/BottomPanel/Debugger/views/Summary.tsx
@@ -37,6 +37,10 @@ export default class Summary extends React.Component<Props> {
       return null
     }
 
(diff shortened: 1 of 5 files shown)


packages/studio-ui/src/web/components/Layout/BottomPanel/Debugger/views/Summary.tsx
```diff
@@ -37,6 +37,10 @@ export default class Summary extends React.Component<Props> {
       return null
     }
 
+    if (!this.props.event.processing?.completed) {
```
franklevasseur, line 40: **this is the fix**

everything else is about fixing react errors in chrome console or adding typings

EFF (approved):
good job 👍

PR #259, opened 2022-01-31
fix(studio): add the nat ext path

allardy:
On the server, you can override the native extensions path to provide os-specific bindings, but it's not present on the studio. this change will cause no issue with existing installations, but will bring back the previous support for os we don't support out of the box.

### DIFF
--- a/packages/studio-be/src/sdk/rewire.ts
+++ b/packages/studio-be/src/sdk/rewire.ts
@@ -8,7 +8,9 @@ const originalRequire = Module.prototype.require
 const platformFolders: string[] = []
 const nativeBindingsPaths: string[] = []
 
-const nativeExBaseFolder = syspath.resolve(require.resolve('@botpress/native-extensions'), '../../bin')
+const nativeExBaseFolder =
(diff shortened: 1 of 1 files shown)


EFF:
Does this really fix the issue ? Had a quick talk to @sebburon and I'm not sure this fully fixes it or not

laurentlp:
Also, isn't putting those bindings inside a package an unnecessary overhead?

And as a side note, we don't drag any bindings in messaging. We should maybe investigate doing the same thing in the studio. I don't see why we couldn't achieve the same result.

allardy:
> Does this really fix the issue ? Had a quick talk to @sebburon and I'm not sure this fully fixes it or not.
> 
> Let me know

Yes, it should fix it for now. The server starts correctly with that env var (because bindings were compiled for RHEL), so with this change, it will also do the same for the studio.

@laurentlp I agree, but it wouldn't fix the issue here. Client uses an OS which is not built for (we have windows, mac and ubuntu). Other OS must compile bindings and provide them

samuelmasse:
@allardy closing since as far as I can tell this isn't needed anymore. @sebburon is this resolved?

sebburon:
Currently being tested since it was opened. I haven't got any feedback since then. I think it's safe to close for now.

PR #262, opened 2022-02-02
fix(ui): zoom in and out from cursor position

rob-gordon:
This fixes the odd zooming behavior in the studio
react-storm-diagram had some errors in how it processed non-integer zoom levels

To create a smooth zoom we 
- store our own floating point zoom value and only pass the rounded version to react-storm
- we control the version we pass by hijacking the `wheel` handler that manages zooming in and out
- zoom imperatively rather than effectually – this just means that we're zooming first and then updating the zoom level we keep in redux state. trying to zoom as an effect of the redux state also may have been getting react-storm out of sync. As a result, there's a few actions and reducers we no longer needed

I changed the minimum zoom to 25 rather than 10 because 10 seemed comically small and 25 was the smallest value in out `select` in the interface. I also set a reasonable maximum (again, from the select). Happy to change or remove this functionality if people see fit.

**Before**

https://user-images.githubusercontent.com/159949/152232211-0669be00-d892-4741-abed-6f9185358179.mov

**After**

https://user-images.githubusercontent.com/159949/152228320-5485e9cd-a100-46a4-88ae-1f2c8c65e13d.mov


laurentlp (commented):
Nice work! Really good cleanup!

I have found single small issue, but otherwise it looks fine to me


https://user-images.githubusercontent.com/9640576/152259302-b6eac085-8425-49cd-be6d-3f405721886b.mp4

laurentlp (approved):
Tested and working perfectly! Awesome work as usual!

PR #265, opened 2022-02-03
fix(builtin): bug-erase-variable studio change

Raid55:
this is a clone of a change made on `botpress/botpress` for the next changes.


samuelmasse:
Hey @Raid55 the conventional commit format uses `:` not `;` ! So it should be fix(builtin): bug-erase-variable studio change

laurentlp (approved):

PR #279, opened 2022-02-07
chore(deps): update dependencies

laurentlp:
This PR aims at lowering the number of security vulnerabilities that we have in our dependencies.

Most of the vulnerabilities are in our `devDependencies` but some we also in the production ones.


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "build": "yarn workspaces foreach -ptiv run build",
     "watch": "yarn workspaces foreach -piv run watch",
-    "clean": "yarn workspaces foreach -piv run clean",
+    "clean": "yarn workspaces foreach -pi run clean",
```
laurentlp, line 11: No need to display info in console (v is for verbose)

packages/studio-ui/tsconfig.json
```diff
@@ -21,5 +21,5 @@
     }
   },
   "include": ["src/**/*.ts", "src/**/*.tsx"],
-  "exclude": ["node_modules", "build", "dist", "example", "rollup.config.js", "webpack.config.js", "**/*.test.ts"]
```
laurentlp, line 24: This list of paths excluded was unnecessary as we only include files inside the src dir.

samuelmasse (approved):
LGTM

PR #282, opened 2022-02-08
chore(layout): use bp3 dropdown for lang switcher

EFF:
rewrite of language switcher with common dropdown.  I'm aware style isn't exactly the same but I believe it's good enough as it touches a tiny portion of the user base.

If you guys don't agree, you are welcome on styling this thing.

@sebburon should fix your issue

Closes https://github.com/botpress/botpress/issues/11405

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/StatusBar/LangSwitcher.tsx
+++ b/packages/studio-ui/src/web/components/Layout/StatusBar/LangSwitcher.tsx
@@ -1,24 +1,15 @@
-import { lang, utils } from 'botpress/shared'
+import { Button } from '@blueprintjs/core'
+import { Dropdown } from 'botpress/shared'
 import _ from 'lodash'
-import React, { Fragment } from 'react'
-import { Dropdown } from 'react-bootstrap'
+import React, { useEffect } from 'react'
 
 import withLanguage from '../../Util/withLanguage'
 
-import ActionItem from './ActionItem'
-import style from './style.scss'
-
-const requireFlag = (code) => {
-  try {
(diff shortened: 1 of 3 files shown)


sebburon:
Just tested and it looks great and fixes [my issue ](https://github.com/botpress/botpress/issues/11405)

packages/studio-ui/src/web/components/Layout/StatusBar/LangSwitcher.tsx
```diff
@@ -1,122 +1,65 @@
...
 
-class LangSwitcher extends React.Component<Props> {
-  elems = {}
-
-  componentDidMount() {
-    this.restoreLastLanguage()
-  }
+export const LanguageSwitcher = (props: Props) => {
+  useEffect(() => {
+    restoreLastLanguage()
```
rob-gordon, line 22: I don't think you need the `restoreLastLanguage` function– you can just move the contents of the function inside of this useEffect

rob-gordon (approved):
Love that you converted the class component to a functional one. Small suggestion on code simplicity but not required. LGTM 👍🏻

packages/studio-ui/src/web/components/Layout/StatusBar/LangSwitcher.tsx
EFF: done ✅

EFF:
@allardy Let's agree to drop the keyboard shortcut I'm pretty sure I'm the only one that uses the keyboard shortcut.

sebburon (approved):
Looking good

allardy:
@EFF totally agree

PR #283, opened 2022-02-08
fix(studio): content picker broken

allardy:
The content picker was broken since https://github.com/botpress/studio/pull/279

image


laurentlp (approved):
Right! Good catch!

laurentlp (approved):

PR #284, opened 2022-02-08
revert(deps): downgrade fs-extra to v9

laurentlp:
This PR fixes an issue with the binary produced with `pkg`. Since pkg uses fs-extra v8, it conflicted, when running the binary, with the version (v10) that was installed in the `studio-be` package.

Related to: https://github.com/vercel/pkg/issues/1180

Closes BUS-214


samuelmasse (approved):

daehli:
Have you updated the next Branch? With this change ?

laurentlp:
> Have you updated the next Branch? With this change ?

@daehli Yes! Let me know if you still have issues.

PR #286, opened 2022-02-09
chore: get rid of topics dead code

EFF:


packages/studio-ui/src/web/views/FlowBuilder/inspector/index.tsx
```diff
@@ -108,19 +107,6 @@ class Inspector extends Component<Props> {
...
-        <SaySomethingForm
-          onDeleteSelectedElements={onDeleteSelectedElements}
-          contentType={currentFlowNode.content?.contentType}
-          formData={currentFlowNode.content?.formData}
-          updateNode={updateNodeAndRefresh}
-          readOnly={readOnly}
-          subflows={subflows}
-        />
-      )
-    }
```
allardy, line 122: Do we get rid of it? You can actually use that node in the current version of Botpress by setting "experimental" to true in your botpress config file. 
image

allardy (commented):
Looks good, just not entirely sure about the say something form. It's actually usable right now, in experimental mode. Either way i'm fine with it, just letting you know that it's not really "dead code"

packages/studio-ui/src/web/views/FlowBuilder/inspector/index.tsx
EFF: Ok we'll decide today what we do with those. Either we make em enabled by default or we remove all of them. I'll change the pr to only topics stuff

EFF: @allardy ✅ 
Brought back the say something node to ensure non breaking change. We'll probably get rid of all content form in 13 and get rid of the experimental thing

PR #296, opened 2022-02-16
chore(deps): upgrade pkg version to sign binary

ptrckbp:
This issue https://github.com/vercel/pkg/commit/0b55f9adba1f52d2aed171ca902d44fffa114af0 is preventing me from including the binary inside the electron app.

I tried packaging, and placing the studio binary within the master branch of botpress/botpress and it works flawlessly, but that's just one environment (Intel Macos). Please let me know if you need me to test this somehow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "eslint-plugin-jsdoc": "37.8.0",
     "jest": "^27.5.0",
     "minimist": "^1.2.5",
-    "pkg": "^4.3.7",
+    "pkg": "^5.5.2",
     "prettier": "^2.5.1",
     "semver": "^7.3.5",
     "ts-jest": "^27.1.3",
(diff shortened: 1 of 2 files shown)


allardy:
Updating pkg means that we also upgrade to a new node version, which also means that there's a bunch of tests that we must do to ensure everything is stable, and we also need to update the version on the main repository at the same time.

And we need to test all binaries (windows, linux, mac)

laurentlp (changes requested):
`pkg` uses `pkg-fetch` to download node. v5.5.2 of pkg depends on v3.2.6 of pkg-fetch which in turn provides NodeJS v12.22.9 (https://github.com/vercel/pkg-fetch/blob/v3.2.6/patches/node.v12.22.9.cpp.patch). All this to say that we also need to update the version of node we use in the .nvmrc file, the CI, and typings (@types/node).

samuelmasse:
@ptrckbp Closing for inactivity. Feel free to re-open

PR #297, opened 2022-02-16
fix(superinput): various ux bugs

Raid55:
- [x] Entering first key, cursor shifts to the left.
- [x] on autofocus, cursor shows up at start of string instead of end
- [x] eval pannel's text overflows past border when too much text


laurentlp (approved):
LGTM

PR #298, opened 2022-02-17
fix(build): fix package on windows & cmd

allardy:
Packaging failed on windows because of the command not including cross-env.

image

Also made a small adjustment to the package.json file, scripts should do a single task... Package shouldn't clean/build/package, it just packages. If those commands must be used together, then they must be typed independently

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "clean": "yarn workspaces foreach -pi run clean",
     "start": "ts-node -T --dir scripts start_studio",
     "setup": "ts-node -T --dir scripts create_default_env --location ./.env.debug",
-    "package": "yarn clean && yarn build && ts-node -T --dir scripts package",
+    "package": "ts-node -T --dir scripts package",
     "release": "ts-node -T --dir scripts release",
     "prettier": "prettier --check './packages/**/*.ts(x)' '!**/*.d.ts'",
     "eslint": "eslint packages/ --ext .ts,.tsx",
--- a/scripts/package.ts
+++ b/scripts/package.ts
@@ -10,7 +10,7 @@ const installBindings = async () => {
 
   for (const platform of platforms) {
     await execute(
-      `./node_modules/.bin/node-pre-gyp install --directory=./node_modules/sqlite3 --target_platform=${platform} --target_arch=x64`,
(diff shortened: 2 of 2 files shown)


scripts/package.ts
```diff
@@ -10,7 +10,7 @@ const installBindings = async () => {
 
   for (const platform of platforms) {
     await execute(
-      `./node_modules/.bin/node-pre-gyp install --directory=./node_modules/sqlite3 --target_platform=${platform} --target_arch=x64`,
+      `cross-env ./node_modules/.bin/node-pre-gyp install --directory=./node_modules/sqlite3 --target_platform=${platform} --target_arch=x64`,
```
laurentlp, line 13: Pretty sure that calling cross-env like this requires having it installed globally.

package.json
```diff
@@ -11,7 +11,7 @@
     "clean": "yarn workspaces foreach -pi run clean",
     "start": "ts-node -T --dir scripts start_studio",
     "setup": "ts-node -T --dir scripts create_default_env --location ./.env.debug",
-    "package": "yarn clean && yarn build && ts-node -T --dir scripts package",
+    "package": "ts-node -T --dir scripts package",
```
laurentlp, line 14: Why remove the clean and build part here? It makes more sense to remove the old dist and then re-build the project to be always sure the resulting bundle only contains the right files.

scripts/package.ts
laurentlp: Also, since we don´t mention those bindings in the assets of pkg (in the main package.json), I think we use the bindings provided by the [native-extensions](https://github.com/botpress/studio/tree/master/packages/native-extensions) package instead.

allardy: Nope, I don't have it installed globally. Dot commands doesn't work on windows. Using cross-env is the only safe option to run commands across all OS

package.json
allardy: Because when I type the command "package", I expect it to just package what is there. If we need to clean and build beforehand, then we can type those commands in succession. Typing a command shouldn't be a "surprise". When you build, it builds, when you clean, it cleans. 

We added gulp in the main repository to handle special commands like `yarn cmd build:core` so we can add commands without denaturing the "official" commands (when you yarn build, it builds everything, and we keep the scripts list small)

samuelmasse (approved):

PR #304, opened 2022-02-23
fix(content): added translations for card buttons

EFF:
Since the json schema is generated for each new chatbot in its content-type folder, we would need a file migration in order to get that fix for existing chatbots. If we want to go that route this needs to be addressed in another pull request. Meanwhile these little changes fixes it for newly created chatbots.
Before: 
Screen Shot 2022-02-22 at 9 40 38 PM

After
Screen Shot 2022-02-22 at 9 37 41 PM


laurentlp (approved):

allardy (approved):
LGTM

PR #305, opened 2022-02-23
chore: content form enhancements

EFF:
I mainly did 3 things here. Take them as a suggestion, I was simply tired of the huge turquoise colors.
I also didn't re-check any logic, my attention was strictly ui. 

1- moved content form submit and cancel buttons in the actual form implementation
2- changed these buttons for blueprint buttons
3- changed the CreateOrEdit modal langage buttons for blueprint buttons

Before: 
Screen Shot 2022-02-22 at 11 08 28 PM

After:
Screen Shot 2022-02-22 at 11 03 38 PM


allardy:
I'd like to test it but somehow the build fails on next
image

rob-gordon:
@EFF I found what may be a regression, though I haven't diagnosed why. If this is only happening for me, then I apologize! I went back and forth between `f_content-form` and `next` and I don't see this issue on next but I do see it on this branch.

Opening the node inspector appears blank because the `sidePanel_hidden` class isn't being removed from the ancestor.

https://user-images.githubusercontent.com/159949/155392338-4c18292c-cb7b-4013-b5f9-c2338a4dbaf2.mov

EFF:
@rob-gordon I had this a few days ago, but it was on next. It's nothing caused by this branch in particular. I will address this in another PR

rob-gordon:
@EFF In that case, looks good to me.

rob-gordon (approved):

PR #315, opened 2022-03-14
Implement text field as standalone

rob-gordon:
This PR
- creates an `Input` component which is just a BP3 re-export
- Adds storybook stories for common Input usage
- Adds a bit of functionality to `SuperInput` so that it supports some of the same functionality as the `Input`; specifically `leftIcon` and `rightElement` props
- Inherits more styles from BP3 via classes and removes the need for some manual CSS'ing we we're doing before


Raid55 (approved):
lgtm!

Raid55 (approved):

PR #316, opened 2022-03-14
fix(studio): remove skill unusable on cloud

allardy:
That skill can't be used on the cloud and is useless for now, so we're removing it.


laurentlp (approved):
LGTM

PR #323, opened 2022-03-25
feat(node-editor): added blank route for node editor

Raid55:


laurentlp:
@Raid55 make sure to use a conventional commit format for your PR name

PR #325, opened 2022-03-28
feat(studio): add download modelweights step before uploading chatbot to cloud

laurentlp:
This PR adds the prerequisites steps to make sure that the models are uploaded with the chatbot when deploying to the cloud.

### To test this feature, you have two options:

1. Use the cloud NLU
    - No configuration is required
    - It will succeed and you will be able to see the model(s) on the disk under `data/<botId>/models/<modelId>`
3. Run the NLU locally 
    - Clone the NLU repository and checkout the `cloud` branch
    - Set `CLOUD_NLU_ENDPOINT` to `http://localhost3200` on Botpress side
    - Create a configuration file for the NLU server by running `yarn start init`
    - Edit the `nlu.config.json` file by adding the config `"modelTransferEnabled": true`
    - Start the NLU server and try to deploy a bot. It will fail but you will be able to see the model(s) on the disk under `data/<botId>/models/<modelId>`

### DIFF
--- a/packages/studio-be/package.json
+++ b/packages/studio-be/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@botpress/native-extensions": "*",
-    "@botpress/nlu-client": "1.0.0-rc.2",
+    "@botpress/nlu-client": "1.0.2-rc.6",
     "axios": "^0.25.0",
     "bluebird": "^3.7.2",
     "bluebird-global": "^1.0.1",
--- a/packages/studio-be/src/core/bots/bot-service.ts
+++ b/packages/studio-be/src/core/bots/bot-service.ts
@@ -219,15 +219,17 @@ export class BotService {
     }
   }
 
-  async exportBot(botId: string): Promise<Buffer> {
+  async exportBot(botId: string, { cloud }: { cloud: boolean } = { cloud: false }): Promise<Buffer> {
     const replaceContent: ReplaceContent = {
       from: [new RegExp(`/bots/${botId}/`, 'g')],
       to: [BOT_ID_PLACEHOLDER]
     }
+    const excludes = ['libraries/node_modules/**/*']
+    if (!cloud) {
+      excludes.push('models/**/*')
+    }
 
-    return this.ghostService
-      .forBot(botId)
-      .exportToArchiveBuffer(['models/**/*', 'libraries/node_modules/**/*'], replaceContent)
+    return this.ghostService.forBot(botId).exportToArchiveBuffer(excludes, replaceContent)
   }
 
   async duplicateBot(sourceBotId: string, destBotId: string, overwriteDest: boolean = false) {
--- a/packages/studio-be/src/studio/cloud/cloud-router.ts
+++ b/packages/studio-be/src/studio/cloud/cloud-router.ts
@@ -1,6 +1,6 @@
 import axios from 'axios'
 import { CloudConfig } from 'botpress/sdk'
-import FormData from 'form-data'
+import FormData from 'form-data'q
 import qs from 'querystring'
 import { StudioServices } from 'studio/studio-router'
 import { CustomStudioRouter } from 'studio/utils/custom-studio-router'
@@ -94,7 +94,9 @@ export class CloudRouter extends CustomStudioRouter {
           return res.status(404)
         }
 
-        const botBlob = await this.botService.exportBot(botId)
+        await this.nluService.downloadAndSaveModelWeights(botId)
+
+        const botBlob = await this.botService.exportBot(botId, { cloud: true })
 
         const botMultipart = new FormData()
         botMultipart.append('botId', introspect.botId)
--- a/packages/studio-be/src/studio/cloud/cloud-router.ts
+++ b/packages/studio-be/src/studio/cloud/cloud-router.ts
@@ -1,6 +1,6 @@
 import axios from 'axios'
 import { CloudConfig } from 'botpress/sdk'
-import FormData from 'form-data'q
+import FormData from 'form-data'
 import qs from 'querystring'
 import { StudioServices } from 'studio/studio-router'
 import { CustomStudioRouter } from 'studio/utils/custom-studio-router'
--- a/packages/studio-be/src/studio/nlu/bot/index.ts
+++ b/packages/studio-be/src/studio/nlu/bot/index.ts
@@ -23,7 +23,7 @@ export class Bot {
   constructor(
     botDef: BotDefinition,
     _configResolver: ConfigResolver,
-    _nluClient: NLUClient,
+    private _nluClient: NLUClient,
     private _defRepo: DefinitionsRepository,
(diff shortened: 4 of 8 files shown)


packages/studio-be/src/core/bots/bot-service.ts
```diff
@@ -219,15 +219,17 @@ export class BotService {
...
 
-  async exportBot(botId: string): Promise<Buffer> {
+  async exportBot(botId: string, { cloud }: { cloud: boolean } = { cloud: false }): Promise<Buffer> {
     const replaceContent: ReplaceContent = {
       from: [new RegExp(`/bots/${botId}/`, 'g')],
       to: [BOT_ID_PLACEHOLDER]
     }
+    const excludes = ['libraries/node_modules/**/*']
+    if (!cloud) {
+      excludes.push('models/**/*')
```
laurentlp, line 229: If cloud is set to true, we don't exclude the models as we want them in the archive.

packages/studio-be/src/studio/nlu/nlu-client/cloud/client.ts
```diff
@@ -40,6 +40,12 @@ export class CloudClient extends StanClient {
 
     this.axios.interceptors.request.use(this._requestInterceptor(tokenCache).bind(this) as any)
     this.axios.interceptors.response.use(undefined, this._errorInterceptor(this.axios as any, tokenCache).bind(this))
+
+    this.modelWeights.axios.interceptors.request.use(this._requestInterceptor(tokenCache).bind(this) as any)
+    this.modelWeights.axios.interceptors.response.use(
+      undefined,
+      this._errorInterceptor(this.axios as any, tokenCache).bind(this)
+    )
```
laurentlp, line 48: Since the `ModelTransferClient` does not use the same `axios` instance as the `NLUClient`, we have to set the interceptors twice (another option would be to refactor the NLU client so clients share the same axios instance). This does not really impact performances in any way since the interceptor's handlers use the same cache.

packages/studio-be/src/studio/nlu/nlu-client/index.ts
```diff
@@ -77,6 +77,15 @@ export class NLUClient {
     }
   }
 
+  public async downloadModelWeights(appId: string, modelId: string): Promise<Buffer> {
+    const downloadRes = await this._client.modelWeights.download(appId, modelId, { responseType: 'arraybuffer' })
+    if (downloadRes.status !== 'OK') {
+      throw new Error(`Download weights received status ${downloadRes.status}`)
```
laurentlp, line 83: I don't know what would be a better error than the one suggested in the client's README. Any idea @franklevasseur ?

packages/studio-be/src/studio/nlu/nlu-service.ts
```diff
@@ -163,6 +163,29 @@ export class NLUService {
...
+    }
+
+    const botConfig = await this.configProvider.getBotConfig(botId)
+
+    if (!botConfig.nluModels) {
+      throw new Error('Missing NLU models. Bot is not trained.')
+    }
+
+    const modelsFolder = 'models'
+    await this.ghost.forBot(botId).deleteFolder(modelsFolder)
```
laurentlp, line 179: We delete the models folder before saving them, in case there was some already.

packages/studio-be/src/studio/nlu/nlu-service.ts
```diff
@@ -163,6 +163,29 @@ export class NLUService {
...
+
+    for (const lang of Object.keys(botConfig.nluModels)) {
+      const modelId = botConfig.nluModels[lang]
+      const modelWeights = await bot.downloadModelWeights(botId, modelId)
+
+      await this.ghost.forBot(botId).upsertFile(modelsFolder, `${modelId}.model`, modelWeights)
```
laurentlp, line 185: We could probably skip the part where we save the models locally, but this would require multiple refactors, so that's why I opted for the easiest path.

packages/studio-be/src/studio/nlu/bot/index.ts
```diff
@@ -76,6 +76,10 @@ export class Bot {
     }
   }
 
+  public async downloadModelWeights(appId: string, modelId: string) {
```
tomkcey, line 79: **nitpick (non-blocking)**: if you could add a return type; i'm not sure what that NLU function returns

tomkcey (commented):
It looks good to me, but I'm not familiar with that codebase so maybe another set of eyes would be beneficial.

packages/studio-be/src/studio/nlu/bot/index.ts
laurentlp: Done

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -94,7 +94,9 @@ export class CloudRouter extends CustomStudioRouter {
           return res.status(404)
         }
 
-        const botBlob = await this.botService.exportBot(botId)
+        await this.nluService.downloadAndSaveModelWeights(botId)
```
franklevasseur, line 97: :+1:

packages/studio-be/src/studio/nlu/bot/index.ts
```diff
@@ -23,7 +23,7 @@ export class Bot {
   constructor(
     botDef: BotDefinition,
     _configResolver: ConfigResolver,
-    _nluClient: NLUClient,
+    private _nluClient: NLUClient,
```
franklevasseur, line 26: 😍

packages/studio-be/src/studio/nlu/nlu-client/cloud/client.ts
franklevasseur: this is fine

packages/studio-be/src/studio/nlu/nlu-client/index.ts
franklevasseur: this is good

packages/studio-be/src/studio/nlu/nlu-service.ts
franklevasseur: :+1:

packages/studio-be/src/studio/nlu/nlu-service.ts
franklevasseur: :+1:

packages/studio-be/src/core/bots/bot-service.ts
franklevasseur: :+1:

packages/studio-be/package.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@botpress/native-extensions": "*",
-    "@botpress/nlu-client": "1.0.0-rc.2",
+    "@botpress/nlu-client": "1.0.2-rc.6",
```
franklevasseur, line 17: 😍

franklevasseur (approved):
I didnt test but the implementation looks good to me

👍

PR #327, opened 2022-03-30
fix(debugger): do not escape user input in debugger logs

laurentlp:
This PR adds the logic to handle displaying raw user data when possible. 

The thumb rule is basic, don't escape characters that aren't escaped by the channel-web or the console.

**Before**

Screenshot from 2022-03-30 18-26-22

**After**

 
Screenshot from 2022-03-30 18-25-28

Closes https://github.com/botpress/botpress/issues/11271

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/BottomPanel/Logs/index.tsx
+++ b/packages/studio-ui/src/web/components/Layout/BottomPanel/Logs/index.tsx
@@ -21,6 +21,9 @@ import logStyle from './style.scss'
 const INITIAL_LOGS_LIMIT = 200
 const MAX_LOGS_LIMIT = 500
 
+const isGlobalLog = (scope: string) => 'logs::*' === scope
+const isCurrentBotLog = (scope: string) => `logs::${window.BOT_ID}` === scope
+
 interface Props {
   toggleBottomPanel: () => void
   emulatorOpen: boolean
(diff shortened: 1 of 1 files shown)


packages/studio-ui/src/web/components/Layout/BottomPanel/Logs/index.tsx
```diff
@@ -107,14 +115,27 @@ class BottomPanel extends React.Component<Props, State> {
...
 
+  renderLogContent(content: string) {
+    const isHTML = (str: string) => {
+      const doc = new DOMParser().parseFromString(str, 'text/html')
+      return Array.from(doc.body.childNodes).some((node) => node.nodeType === 1)
+    }
+
+    if (isHTML(content)) {
+      return <span className={logStyle.message} dangerouslySetInnerHTML={{ __html: content }} />
+    } else {
```
samuelmasse, line 126: Why do we want do dangerously do anything with user provided input? Doesn't that make use vulnerable to XSS

laurentlp: I personally have no clue why but since the code was there before, we can consider that this PR does not bring any vulnerabilities.

samuelmasse: I think this is a major vulnerability. Someone in the webchat could send a script tag as a message, then some server admin could check the logs and have a random script executing in their browser. This script could steal their login token and give root access to the hacker. Who's asking for this feature?

sebburon (approved):
Tested and good

PR #329, opened 2022-03-31
chore(core): use proper env var for controller endpoint

laurentlp:
This PR makes sure to bring changes made here https://github.com/botpress/botpress/pull/11676 to the studio


samuelmasse (approved):
LGTM

PR #330, opened 2022-04-04
fix(dev): local cdm setup

samuelmasse:
Closes DEV-2550

### DIFF
--- a/packages/studio-be/src/studio/cloud/cloud-router.ts
+++ b/packages/studio-be/src/studio/cloud/cloud-router.ts
@@ -11,11 +11,11 @@ export class CloudRouter extends CustomStudioRouter {
   }
 
   getBearerToken(cloud: CloudConfig) {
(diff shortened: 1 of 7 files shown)


packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -11,11 +11,11 @@ export class CloudRouter extends CustomStudioRouter {
   }
 
   getBearerToken(cloud: CloudConfig) {
-    const { oauthUrl, clientId, clientSecret } = cloud
+    const { clientId, clientSecret } = cloud
 
     return axios
       .post(
-        oauthUrl,
+        process.CLOUD_OAUTH_ENDPOINT!,
```
laurentlp, line 18: Why the non-null assertion operator?

samuelmasse: It's a random leftover I removed it

PR #334, opened 2022-04-06
fix(scope): call oauth with nlu scope

tomkcey:


anthonyvallee:
Tested. Works when calling the Hydra public endpoint. The `nlu` scope is added by the Cloud on client creation, the Studio that scope in the token.

```
$ curl -X POST \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d"client_id=<client_id>&client_secret=<client_secret>&grant_type=client_credentials&scope=nlu" \
  http://127.0.0.1:4444/oauth2/token

{"access_token":"<redacted>","expires_in":3599,"scope":"nlu","token_type":"bearer"}
```

EFF (approved):

PR #339, opened 2022-04-14
fix(cloud): cloud router error handling

EFF:
Ok this seems like a lot but its majorly moving code around in some private functions and provide better error handling when trying to deploy a chatbot to bp cloud.

Essentially if you look at the original code, request were never closed in the advent of an error. `res.status(400)` will not send a response. I simply replace the sending of status code by throwing our standard http errors and let the asyncMiddleware do its job.

The end result is pretty simple, when everything is fine, the chatbot gets deployed.

Screen Shot 2022-04-13 at 11 39 38 PM

When there is an error, a proper error message is displayed in the ui. Ex: 
Screen Shot 2022-04-13 at 5 09 32 PM
Screen Shot 2022-04-13 at 5 11 45 PM
Screen Shot 2022-04-13 at 5 10 38 PM

Ui also renders the error message provided in the response body if present, in the case of an unexpected error see custom here : 
Screen Shot 2022-04-13 at 4 46 50 PM

In the end this will provide much better information to the end user and support.
Notice that I also added proper logging which avoids a support nightmare.

closes BUS-334


tomkcey (approved):
LGTM

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -25,19 +50,103 @@ export class CloudRouter extends CustomStudioRouter {
           scope: 'nlu'
         })
       )
-      .then((res) => `Bearer ${res.data?.access_token}`)
-      .catch((err) => null)
+
+      if (!data.access_token) {
```
laurentlp, line 54: So if I understand correctly, it's possible that the call to the OAuth endpoint returns a 2XX code without giving an access token?

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -25,19 +50,103 @@ export class CloudRouter extends CustomStudioRouter {
...
+        throw new UnauthorizedError('Cloud keys are invalid')
+      }
+
+      return `Bearer ${data?.access_token}`
+    } catch (err) {
+      const message = `Error fetching cloud bearer token using cloud config. clientId: ${clientId}, clientSecret: ${
+        clientSecret === undefined ? clientSecret : clientSecret.replace(/./g, '*')
```
laurentlp, line 61: How can the client token be undefined? Is it just to be extra safe?

```suggestion
        clientSecret?.replace(/./g, '*')
```

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -25,19 +50,103 @@ export class CloudRouter extends CustomStudioRouter {
...
+        `${process.CLOUD_CONTROLLER_ENDPOINT}/v1/introspect`,
+        this.getCloudAxiosConfig(bearerToken)
+      )
+      if (!introspectData.botId) {
+        throw new NotFoundError("Cloud chatbot doesn't exist")
+      }
+
+      return introspectData as Introspect
+    } catch (err) {
+      const message = 'Error fetching cloud intropsect endpoint'
```
laurentlp, line 80:
```suggestion
      const message = 'Error fetching data from the cloud introspect endpoint'
```

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -25,19 +50,103 @@ export class CloudRouter extends CustomStudioRouter {
...
+  private async getCloudBotInfo(cloudBotId: string, bearerToken: string): Promise<string> {
+    try {
+      const { data } = await axios.get(
+        `${process.CLOUD_CONTROLLER_ENDPOINT}/v1/bots/${cloudBotId}`,
+        this.getCloudAxiosConfig(bearerToken)
+      )
+
+      return data
+    } catch (err) {
+      const defaultMessage = `Error fetching cloud bot info. Cloub bot id ${cloudBotId}`
```
laurentlp, line 95:
```suggestion
      const defaultMessage = `Error fetching cloud bot info. Cloud bot id ${cloudBotId}`
```

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -25,19 +50,103 @@ export class CloudRouter extends CustomStudioRouter {
...
+  }
+
+  private async getCloudBotInfo(cloudBotId: string, bearerToken: string): Promise<string> {
+    try {
+      const { data } = await axios.get(
+        `${process.CLOUD_CONTROLLER_ENDPOINT}/v1/bots/${cloudBotId}`,
+        this.getCloudAxiosConfig(bearerToken)
+      )
+
+      return data
```
laurentlp, line 93: Are you sure that this call doesn't return an object instead of a string?

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -46,94 +155,40 @@ export class CloudRouter extends CustomStudioRouter {
...
-          return
-        }
 
+        const cloud = await this.getCloudBotOauthConfig(botId)
         const bearerToken = await this.getBearerToken(cloud)
-        if (!bearerToken) {
-          return res.status(401)
+        const cloudBotMeta = await this.getIntrospect(bearerToken)
+        if (cloudBotMeta.runtimeStatus !== 'ACTIVE') {
+          const message = `Cloud runtime ${cloudBotMeta.runtimeName} for cloud bot: ${cloudBotMeta.botId}, ${cloudBotMeta.botName} isn't ready.`
```
laurentlp, line 179:
```suggestion
          const message = `Cloud runtime ${cloudBotMeta.runtimeName} for cloud bot: ${cloudBotMeta.botId} isn't ready.`
```

Or 

```suggestion
          const message = `Cloud runtime ${cloudBotMeta.runtimeName} for cloud bot: ${cloudBotMeta.botName} isn't ready.`
```

packages/studio-be/src/studio/cloud/cloud-router.ts
EFF: that was Already there I didn't change the logic

packages/studio-be/src/studio/cloud/cloud-router.ts
EFF: it was already there. If its undefined I want it to appear "undefined" in the logs so we can know about it

laurentlp:
I get this error when I delete my bot on the CDM and then try to re-deploy it:

I think this error might be misleading since it contains a `:` at the end of the error message, which makes me think there is something more when there isn't:

Screenshot from 2022-04-14 17-13-15

EFF:
@laurentlp weird there's no : 
https://github.com/botpress/studio/blob/610c4ac076ceeb769f32fa975af9ee631276b1ad/packages/studio-be/src/studio/cloud/cloud-router.ts#L81

packages/studio-ui/src/web/components/Layout/Sidebar.tsx
```diff
@@ -44,13 +44,6 @@ const BASIC_MENU_ITEMS = [
     rule: { res: 'module.qna', op: 'write' },
     icon: 'chat'
   },
-  {
-    id: 'libraries',
-    name: lang.tr('libraries.fullName'),
-    path: '/libraries',
-    rule: { res: 'module.code-editor', op: 'read' },
-    icon: 'book'
-  },
```
laurentlp, line 53: So we disable it for both cloud and non cloud bots?

packages/studio-be/src/studio/cloud/cloud-router.ts
```diff
@@ -9,6 +9,7 @@ import { StudioServices } from 'studio/studio-router'
 import { CustomStudioRouter } from 'studio/utils/custom-studio-router'
 
 const MAX_BODY_CLOUD_BOT_SIZE = 100 * 1024 * 1024 // 100 MB
+const TOO_LARGE_MESSAGE = 'Chatbot is too large to be uploaded on Botpress cloud'
```
laurentlp, line 12: Translation?

packages/studio-ui/src/web/components/Layout/Sidebar.tsx
EFF: yes, for this particular release. It is still accessible through url though

packages/studio-be/src/studio/cloud/cloud-router.ts
EFF: That's an error message, we never translate them

PR #356, opened 2022-07-05
chore(ui): add support for all iso language code

EFF:
I simply use a library instead of managing all language code manually. Notice that I "hijacked" the `tr` function so it is supported everywhere we use it without breaking the usage.

We can also make the same change in the admin code at some point

Little sign of ❤️ to @sebburon

### DIFF
--- a/packages/ui-shared/src/translations/index.tsx
+++ b/packages/ui-shared/src/translations/index.tsx
@@ -1,6 +1,9 @@
+import ISOLanguages from '@cospired/i18n-iso-languages'
 import { MultiLangText } from 'botpress/sdk'
 import { isEmpty, merge } from 'lodash'
 import { createIntl, createIntlCache, IntlShape } from 'react-intl'
+
+const ISO_LANG_PREFIX = 'isoLangs.'
 
 import en from './en.json'
 import es from './es.json'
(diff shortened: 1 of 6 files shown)


packages/ui-shared/src/translations/index.tsx
```diff
@@ -102,6 +112,12 @@ const lang = (id: string | MultiLangText, values?: { [variable: string]: any }):
     return id[locale] || id[defaultLocale] || ''
   }
 
+  // pattern is isolangs.{code}
+  if (id.startsWith(ISO_LANG_PREFIX)) {
+    const code = id.substring(ISO_LANG_PREFIX.length, ISO_LANG_PREFIX.length + 2)
```
laurentlp, line 117: Are you sure the code will always be two chars? Otherwise, you could simply remove the prefix from the string.

laurentlp (approved):
Small comment otherwise LGTM!

packages/ui-shared/src/translations/index.tsx
EFF: yes 100% sure. cannot only remove prefix from string because we sometimes add .name after.

PR #372, opened 2022-07-25
fix(auth): fix authenticating with cookie storage

laurentlp:
This PR fixes an issue where when the JWT token was stored in the cookie storage, the call to `/env` would fail with a 401. The request was denied because the `window` variable '`USE_JWT_COOKIES`' was not set at this point.

To fix the problem, I renamed the `/branding.js` call to `/public-env.js` so that we can send any **public** environment variable to the front-end without authentication and passed the value of `process.USE_JWT_COOKIES` in it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botpress/studio",
-  "version": "0.0.56",
+  "version": "0.0.55",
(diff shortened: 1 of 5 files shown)


sebburon (approved):
Tested and fixes the issue

PR #374, opened 2022-07-27
fix(security): fix access to public envs when bot is unmounted

laurentlp:
This PR fixes an issue where unmounting a bot would prevent the user from accessing the bot settings page.

The issue came from the `checkBotVisibility` middleware that checks whether or not the bot is accessible (mounted). It contains an exception for the `/env` route and placeholder bots (`___`). Meaning that we were missing the exception for the `public-env` route.


sebburon (approved):
LGTM

PR #375, opened 2022-07-27
fix(debugger): fix display of log timestamp

laurentlp:
This PR fixes the display of the date in the studio's debugger.

**before**

Screen Shot 2022-07-27 at 10 33 09 AM

**after**

Screenshot from 2022-07-27 11-24-04


sebburon (approved):
Perfect !

PR #377, opened 2022-08-03
chore(studio): remove enterprise button

allardy:


packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
```diff
@@ -28,15 +28,7 @@ const Toolbar: FC<Props> = (props) => {
...
-        {window.IS_PRO_ENABLED ? (
-          <span />
-        ) : (
-          <Tooltip position="right-bottom" content={lang.tr('studio.flow.toolbar.salesCallToActionDescription')}>
-            <a className={style.cta_btn} target="_blank" href="https://botpress.com/request-trial-from-app">
-              {lang.tr('studio.flow.toolbar.salesCallToAction')}
-            </a>
-          </Tooltip>
-        )}
+        <span />
```
EFF, line 31: why keep an empty span ?

EFF (approved):
lgtm

packages/studio-ui/src/web/components/Layout/Toolbar/index.tsx
allardy: Because of the alignment (space-between). Didn't want to play with the style, and if we add stuff in the future, it will be ready

PR #378, opened 2022-08-03
feat(cloud): Personal Access Tokens

spg:
Handles Personal Access token flow in the Studio

* If PAT is not present in Local Storage, prompt for PAT is displayed. PAT is validated against CDM
* Allows selection of Cloud Workspace
* Trains bot and deploys it to Cloud

See https://linear.app/botpress/issue/BUS-645/support-personal-access-tokens-in-studio

Closes BUS-645


allardy:
Clicking on the button brings me to an invalid page: 
image


image

call to action is a bit unclear, "token invalid" is an "error message" but opens a page

spg:
@allardy 
* i18n fix in 96f7e5b
* the page is not yet deployed in `/settings#pat` page is not yet deployed, depends on https://github.com/botpress/cdm-cloud/pull/475

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const [state, dispatch] = useReducer(reducer, props, init)
+
+  const { status } = state
+
+  useEffect(() => {
+    const ac = new AbortController()
+    const startTrain = async () => {
+      for (const lang of Object.keys(trainSessions)) {
+        await axios.post(`${BASE_NLU_URL}/train/${lang}`, null, { signal: ac.signal })
+      }
```
EFF, line 84: **question** is there any cases where you have a multilingual chatbot that has no train session for one or more languages ? 

In other words, should we use another source to start training in all languages supported by the chatbot ?
Or should we implement a post to /train which would train all languages at once in the backend and handle that logic in the backend

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const Deploy = (props: Props): JSX.Element => {
+  const { pat, trainSessions, workspaceId, onCompleted } = props
+
+  const [state, dispatch] = useReducer(reducer, props, init)
+
+  const { status } = state
+
+  useEffect(() => {
+    const ac = new AbortController()
```
EFF, line 80: **nitpick (non-blocking):** seems like you're using this pattern quite often, you could create a reusable react hook for it instead

packages/studio-ui/src/web/components/Layout/TopNav/cloud/deploy.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { status, text } = props
+  const chars: { [s in Status]: string | JSX.Element } = {
+    pending: '⌛',
+    'in-progress': <Spinner size={SpinnerSize.SMALL} />,
+    completed: '✅',
+    failed: '🚫'
+  }
+
+  return (
+    <div style={{ display: 'flex', justifyContent: 'space-between', gap: '10px' }}>
```
EFF, line 17: as a matter of consistency, move styles in .scss file please 🙏 .
This comments applies to everywhere you write style=

packages/studio-ui/src/web/components/Layout/TopNav/cloud/local-storage-pat.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const LOCALSTORAGE_KEY = 'bp/pat'
+
+export const LocalStoragePat = (props: { onCompleted: (value?: string) => void }): JSX.Element => {
+  const { onCompleted } = props
+
+  useEffect(() => {
+    const initialPatValue = localStorage.getItem(LOCALSTORAGE_KEY)
+
+    const ac = new AbortController()
```
EFF, line 12: **question:** why use a cancellable request everywhere ? is this absolutely necessary ? can it lead to some non-desirable side-effects ? (e.g components gets unmounted but the user still wants chatbot to be deployed)

packages/studio-ui/src/web/components/Layout/TopNav/cloud/local-storage-pat.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect } from 'react'
+import { fetchPatStatus } from './pat'
+
+const LOCALSTORAGE_KEY = 'bp/pat'
+
+export const LocalStoragePat = (props: { onCompleted: (value?: string) => void }): JSX.Element => {
```
EFF, line 6: Why write a react component if it's to only retrieve pat and never render anything ?

packages/studio-ui/src/web/components/Layout/TopNav/cloud/pat-input.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <div>
+      <div className={style.patContainer}>
+        <InputGroup
+          className={style.patInput}
+          placeholder={lang.tr('topNav.deploy.enterPersonalAccessToken')}
+          value={pat}
+          onChange={(e) => onChange(e.target.value)}
+          rightElement={
+            valid ? (
+              <Button disabled={true} icon={'tick-circle'} minimal={true} />
```
EFF, line 26: I don't know if you wanted to be explicit or because you didn't know but you don't have to write `={true}` when you want a boolean prop to be true. in this case 

```html
<Button disabled minimal icon="tick-circle"/>
```

packages/studio-ui/src/web/components/Layout/TopNav/cloud/pat.ts
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios'
+
+export const fetchPatStatus = async (pat: string, ac: AbortController): Promise<boolean> => {
+  const resp = await axios.get(`${window.CLOUD_CONTROLLER_ENDPOINT}/v1/pat/authenticate`, {
+    headers: { Authorization: `bearer ${pat}` },
```
EFF, line 5: **question:** isn't it `Bearer ...` and not `bearer ...` or our auth stack doesn't give about case ?

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
```diff
@@ -0,0 +1,13 @@
+import { NLU } from 'botpress/sdk'
+
+/**
+ * CDM Workspace
+ */
+export interface Workspace {
```
EFF, line 6: **suggestion :**
```ts
export interface BPCloudWorkspace {
  id: string
  name: string
}
```

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: string
+  name: string
+}
+
+export interface TrainSessions {
```
EFF, line 11: pretty sure this type already exist somewhere

packages/studio-ui/src/web/components/Layout/TopNav/cloud/workspace.tsx
```diff
@@ -0,0 +1,43 @@
+import { Button, Intent, MenuItem } from '@blueprintjs/core'
+import { Select } from '@blueprintjs/select'
+import { lang } from 'botpress/shared'
+import React, { useState } from 'react'
+import style from './style.scss'
+import { Workspace } from './types'
+
+const WorkspaceSelect = Select.ofType<Workspace>()
+
+export function WorkspaceSelector(props: {
```
EFF, line 10: we have [this wrapper component](https://github.com/botpress/studio/blob/4ded33437ef9f5f6fe2274dde3f30f56e6c1df39/packages/ui-shared/src/Dropdown/index.tsx#L33-L34) that handles most of the blueprint select boilerplate if that could be of any use

packages/studio-be/src/studio/cloud/schemas.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod'
+
+export const DeployRequestSchema = z.object({
```
EFF, line 3: sadly we use joi everywhere in this project. as a matter of consistency, can you rewrite this schema in with joi (if you have time)

packages/studio-ui/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "useUnknownInCatchVariables": false,
+    "strictNullChecks": true,
```
EFF, line 9: if this rule applies to all files and not only files that are edited can you revert it

EFF (commented):
Most of my comments are nitpicks but a few questions needs to be addressed.

packages/studio-be/src/studio/cloud/cloud-client.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export const MAX_BODY_CLOUD_BOT_SIZE = 100 * 1024 * 1024 // 100 MB
+
+export class CloudClient {
+  constructor(private logger: Logger) { }
+
+  public async createBot(props: {
+    personalAccessToken: string
+    name: string
+    workspaceId: string
```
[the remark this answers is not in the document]
spg: I really like the `props` pattern because 
* it allows the caller to be less careful about the order of arguments. For example, it's easy for the caller to mess up arguments ordering with `func(arg1: string, arg2: string, arg3: string)`, e.g. `func(arg2, arg1, arg3) // Typescript will not complain`, whereas with: `func(props: {arg1: string; arg2: string; arg3: string})` the user is less concerned about the arguments order
* If the user has 3 local variables `arg1, arg2 and arg3`, it's super easy to call the function and have Typescript help you if you mis-assign variables: `func({arg1, arg2, arg3})`, and `func({arg1, arg1, arg3}) // TS error`

packages/studio-ui/src/web/components/Layout/TopNav/cloud/pat-input.tsx
spg: Fixed in 4355c311a19e6fafa42eaa945bf995b3c4b7c65e

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: It's already in the `cloud/` directory, thus we should assume it's cloud-related

packages/studio-ui/tsconfig.json
spg: It applies to all files. I think it's a good idea to adopt it now, as it will help Typescript catch a lot more errors (I know it helped me catch many bugs in my code).

I understand this adds a lot of red everywhere in the codebase, but I think it's mandatory to activate this feature sooner than later

packages/studio-be/src/studio/cloud/schemas.ts
spg: Screen Shot 2022-08-08 at 10 14 54 AM
We use joi in only a handful of files. We can migrate to zod gradually. They can live together peacefully during this period of transition.

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step1-pat.tsx
```diff
@@ -0,0 +1,139 @@
...
+      debouncedValidatePat.cancel()
+    }
+  }, [])
+
+  useEffect(() => {
+    if (status === 'checking_new_pat') {
+      const { newPat } = state
+      void debouncedValidatePat(newPat, debounceAc)
+    }
+  }, [status])
```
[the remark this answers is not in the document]
spg: There's already a save button

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step2-workspace.tsx
```diff
@@ -0,0 +1,111 @@
...
+interface OwnProps {
+  pat: string
+  onCompleted: (selectedWorkspaceId: string) => void
+}
+
+type StateProps = ReturnType<typeof mapStateToProps>
+type DispatchProps = typeof mapDispatchToProps
+type Props = DispatchProps & StateProps & OwnProps
+
+type State =
```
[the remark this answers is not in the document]
spg: I was inspired [by this article](https://www.sumologic.com/blog/react-hook-typescript/), which declares all possible states to help the compiler. IMO this is a really elegant patter that we should try to use more, as it helped be uncover a lot of potential mistakes

packages/studio-be/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "moduleResolution": "node",
     "experimentalDecorators": true,
     "useUnknownInCatchVariables": false,
+    "strictNullChecks": true,
```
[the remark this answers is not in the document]
spg: I think we should leave it on, see https://github.com/botpress/studio/pull/378#discussion_r940292381

packages/studio-be/src/studio/cloud/cloud-client.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  public async uploadBot(props: {
+    personalAccessToken: PersonalAccessToken
+    botId: string
+    botMultipart: FormData
+  }): Promise<void> {
+    const { personalAccessToken, botId, botMultipart } = props
+    const axiosConfig = _.merge(this.getCloudAxiosConfig({ token: personalAccessToken, principals: { botId } }), {
```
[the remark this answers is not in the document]
spg: Will keep lodash, I like how using the `merge` function clarifies the intent

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: I could rename it to `CDMWorkspace` though

packages/studio-ui/src/web/components/Layout/TopNav/cloud/local-storage-pat.tsx
spg: It is needed to avoid memory leaks. If a promise is not settled when a component is unmounted, [React will complain](https://stackoverflow.com/questions/54954385/react-useeffect-causing-cant-perform-a-react-state-update-on-an-unmounted-comp).

Cancelling requests avoid these potential memory leaks

packages/studio-be/src/studio/cloud/schemas.ts
allardy: Joi is used litterally in every codebase related to botpress... I strongly recomment keeping Joi and not add another library that everyone will have to learn when interacting with this code

packages/studio-be/src/core/security/auth-service.ts
```diff
@@ -56,6 +56,10 @@ export class AuthService {
       })
     })
   }
+
+  public async findUser({ email, strategy }: { email: TokenUser['email']; strategy: TokenUser['strategy'] }) {
```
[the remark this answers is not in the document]
spg: I think it's simpler to always ensure the communication chain is `router -> service -> repo`. In other words, the router should not use the repo. I prefer that we're strict about this, in order to make the codebase easier to navigate.

packages/studio-be/src/studio/cloud/cloud-client.ts
```diff
@@ -0,0 +1,125 @@
+import axios from 'axios'
+import { Logger } from 'botpress/sdk'
+import { isCDMError } from 'common/errors'
+import FormData from 'form-data'
+import { constants } from 'http2'
+
+import _ from 'lodash'
+import qs from 'qs'
+import { Result, Ok, Err } from 'ts-results'
```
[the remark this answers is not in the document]
spg: It brings the concept of type-safe errors in our codebase. I think it will help the reliability of our codebase if clients of an API can know in advance all error types returned by a function. More here https://blog.logrocket.com/improve-error-handling-typescript-exhaustive-type-checking/

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: Could not find it anywhere

packages/studio-be/src/studio/cloud/schemas.ts
spg: Not anymore. All new codebases in the company in the last few months are uzing Zod. In a Typescript codebase, zod is clearly better than joi due to it's type inference feature. I think we should move to Zod, and we can do it incrementally.

packages/studio-ui/src/web/components/Layout/TopNav/cloud/pat.ts
spg: it's case insensitive at the [Traefik](https://github.com/botpress/infra/blob/master/cloud/lib/controller-stack.ts#L522) and [CDM](https://github.com/botpress/cdm-cloud/blob/develop/api/src/personal-access-tokens/utils.ts#L12) levels

packages/studio-be/src/studio/cloud/cloud-client.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public async createBot(props: {
+    personalAccessToken: string
+    name: string
+    workspaceId: string
+  }): Promise<Result<Bot, CDMConflictError | UnexpectedError>> {
+    const { personalAccessToken, name, workspaceId } = props
+    try {
+      const { data } = await axios.post(
+        `${process.CLOUD_CONTROLLER_ENDPOINT}/v1/bots`,
```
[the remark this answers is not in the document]
spg: Fixed in 9cf04b5

Unfortunately the PAT changes on each request and therefore cannot be passed as a constructor param, since the `CloudClient` is only constructed once at the server startup

packages/studio-be/src/studio/cloud/cloud-service.ts
```diff
@@ -0,0 +1,139 @@
...
+import { NLUService } from 'studio/nlu'
+import { Result, Ok, Err } from 'ts-results'
+import { CloudClient, MAX_BODY_CLOUD_BOT_SIZE } from './cloud-client'
+import { CDMConflictError, CreateBotError, RuntimeNotActiveError, UnexpectedError } from './errors'
+import { Bot } from './types'
+
+export class CloudService {
+  constructor(private cloudClient: CloudClient, private botService: BotService, private nluService: NLUService) { }
+
+  public async deployBot(props: {
```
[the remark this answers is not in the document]
spg: Done in 955880f

packages/studio-ui/src/web/components/Layout/TopNav/cloud/deploy.tsx
spg: Done in cd999da

packages/studio-ui/src/web/components/Layout/TopNav/cloud/local-storage-pat.tsx
spg: Fixed in 62e10bb

packages/studio-be/src/common/errors.ts
```diff
@@ -0,0 +1,22 @@
+import axios, { AxiosError } from 'axios'
```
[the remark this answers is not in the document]
spg: The errors are in `common/errors.ts` so they can be reused in the `studio-ui` as well

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step2-workspace.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { status } = state
+
+  useEffect(() => {
+    const ac = new AbortController()
+
+    const fetchWorkspaces = async () => {
+      const { data: workspaces } = await axios.get(`${window.CLOUD_CONTROLLER_ENDPOINT}/v1/workspaces`, {
+        signal: ac.signal,
+        headers: { Authorization: `bearer ${pat}` }
+      })
```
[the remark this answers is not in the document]
spg: That would involve moving the `CloudClient` to `common` and ensure it is compatible both in the browser and the backend. I don't think it's worth the effort now, since the browser only does light usage of the controller API. What do you think?

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
spg: These are all valid questions, but I thought they fell out of scope of this PR.
Personally I would prefer a single `/train` endpoint that would train all languages, but this would have impacts outside the scope of this PR (e.g. the status bar)

packages/studio-ui/src/web/components/Authentication/index.jsx
```diff
@@ -25,17 +25,26 @@ const ensureAuthenticated = WrappedComponent => {
...
+        (error) => {
+          if (axios.isCancel(error)) {
+            return Promise.reject(new Error('request cancelled'))
+          }
+          if (axios.isAxiosError(error)) {
+            if (error.response.status !== 401) {
+              return Promise.reject(error)
+            }
+            if (error.response.config.url !== `${window.API_PATH}/admin/ping`) {
+              this.checkAuth() // just to make sure
```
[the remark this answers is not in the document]
spg: Not sure I understand your question

packages/studio-be/src/studio/cloud/types.ts
```diff
@@ -0,0 +1,68 @@
+type RuntimeStatus =
```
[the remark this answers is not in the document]
spg: yes I think it would be a good idea

packages/studio-ui/src/web/components/Layout/TopNav/cloud/workspace.tsx
spg: I would rather not use it, as it's not type safe: https://github.com/botpress/studio/blob/e5bd786a809ffe3cf9c4927022a7be7a3871a975/packages/ui-shared/src/Dropdown/typings.d.ts#L6

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step1-pat.tsx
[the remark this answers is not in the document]
spg: I'll remove all debounce logic

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step2-workspace.tsx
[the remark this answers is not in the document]
spg: I'll try moving the CloudClient to `common`

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
spg: I'll check if TrainSessions include all lang keys, all the time

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
spg: I'll try this
```
function abortController() {
  const ac = new AbortController()
  useEffect(() => {
    return () => {
      // Cancel all pending requests
      ac.abort()
    }
  }, [])
  
  return axios.createInstance({signal: ac})
}

const axios = useAbortedAxios() // axios has
```

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: I'll check if I can use the type directly from the NLU client

packages/studio-be/src/studio/cloud/schemas.ts
spg: I'll revert to joi for now

packages/studio-be/src/studio/cloud/types.ts
```diff
@@ -0,0 +1,68 @@
...
+  cloud: BotCloudConfiguration
+  messaging: BotMessagingConfiguration
+}
+
+interface BotCloudConfiguration {
+  clientId: string
+  clientSecret: string
+}
+
+// TODO: create a type for input messaging config and output. input doesnt have webhook token
```
laurentlp, line 62: What do you mean by input and output here?

packages/studio-be/src/studio/cloud/cloud-service.ts
```diff
@@ -0,0 +1,139 @@
+import { CloudConfig } from 'botpress/sdk'
+import { UnreachableCaseError } from 'common/errors'
+import { BotService } from 'core/bots'
+import { backOff } from 'exponential-backoff'
+import FormData from 'form-data'
+import _ from 'lodash'
+import { NLUService } from 'studio/nlu'
+import { Result, Ok, Err } from 'ts-results'
```
laurentlp: Nice!

packages/studio-be/src/studio/cloud/types.ts
spg: Don't know, I didn't write this comment

packages/studio-be/src/studio/cloud/schemas.ts
spg: Fixed in efea418

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
spg: https://github.com/botpress/studio/blob/df96e6e5be31c690cabff83e1d72b1898f20531d/packages/studio-ui/src/web/components/Layout/StatusBar/TrainingStatus/index.tsx#L100 we can see that the TrainSessions are mapped to the bot's languages. Therefore, TrainSessions languages are the same as the bot's languages

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: Fixed in fb992a7

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step1-pat.tsx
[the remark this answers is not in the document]
spg: Fixed in 07f06383626236a837f58df15f216adc5a7c4558

packages/studio-ui/src/web/components/Layout/TopNav/cloud/types.ts
spg: Using the NLU client's typings would involve modifying the `TrainingSession` interface [from the SDK](https://github.com/botpress/studio/blob/be32f12148736a4a5243ddd1408f761f82f26155/packages/studio-be/src/sdk/botpress.d.ts#L296). I think this falls out of scope of this PR.

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step2-workspace.tsx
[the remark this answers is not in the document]
spg: Fixed in d1c06cc

packages/studio-be/src/common/cloud-client.ts
```diff
@@ -27,12 +29,26 @@ class CloudClientError extends VError {
...
+  }
+
+  public async listWorkspaces(props: {
+    personalAccessToken: PersonalAccessToken
+    signal: AbortSignal
+  }): Promise<CDMWorkspace[]> {
+    const { personalAccessToken, signal } = props
+
+    const config = this.getCloudAxiosConfig({ token: personalAccessToken, principals: {} })
+    const { data: workspaces } = await this.axios.get('/workspaces', {
```
EFF: pretty sure you can type an axios call instead of a hard cast when returning `axios.get<T>`

packages/studio-be/src/studio/cloud/oauth-client.ts
```diff
@@ -0,0 +1,41 @@
...
+    const { clientId, clientSecret } = props
+
+    try {
+      const { data } = await this.axios.post(
+        '/',
+        qs.stringify({
+          client_id: clientId,
+          client_secret: clientSecret,
+          grant_type: 'client_credentials',
+          scope: 'nlu'
```
EFF, line 21: only nlu ?

packages/studio-be/src/common/cloud-client.ts
spg: Fixed in d1bec6fa8d826eae7d659ccf985b47d89117a869

packages/studio-be/src/studio/cloud/oauth-client.ts
spg: Yeah it's been working fine so far with that, we can parametrize it in the future. I'd rather keep it like this for now.

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
EFF: this can be done another time, really not salty about this

EFF (dismissed):
if working fine, good to god ✅

PR #380, opened 2022-08-10
fix(cloud): close popup when upload fails

spg:
Now closes the popup if the upload fails
https://www.loom.com/share/15d204f1068c474c91b2cf0c06ba6785

### DIFF
--- a/packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
+++ b/packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
@@ -48,7 +48,7 @@ function reducer(state: State, action: Action): State {
     case 'upload/ended':
       return { status: 'upload_completed' }
     case 'upload/failed':
-      return { status: 'upload_completed' }
+      return { status: 'upload_failed' }
     default:
       throw new UnreachableCaseError(action)
   }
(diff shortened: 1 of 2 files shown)


packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
```diff
@@ -77,6 +84,8 @@ export const Deploy = (props: Props): JSX.Element => {
       } catch (err) {
         if (axios.isAxiosError(err)) {
           toast.failure(err.response?.data?.message || 'An error occured while activating the bot')
+          dispatch({ type: 'activate/failed' })
```
laurentlp, line 87: Just for my personal understanding what does "activating a bot" means?

laurentlp (commented):
Small comment otherwise LGTM!

packages/studio-ui/src/web/components/Layout/TopNav/cloud/Step3-deploy.tsx
spg: You can find the full logic here https://github.com/botpress/studio/blob/d3f519ac00226d91a8c582f9f4a5c90b52cc7d7d/packages/studio-be/src/studio/cloud/cloud-router.ts#L23

Basically: activating ensures that:
1. the bot is created on the Cloud
2. that the bot is unmounted then remounted on the NLU Service (so that the bot uses the NLU Cloud Client). When a bot is created in the Studio, it uses the local NLU client. We want the bot to switch to the NLU Cloud Client, so that the training happens on the cloud

Once activating is completed, we can then train the bot against the cloud, and deploy it.

Hope this helps!

laurentlp (approved):

PR #382, opened 2022-08-16
feat(studio): preview content elements from skills

davidvitora:
This feature allows the flow generator to return a list of content elements to be previewed in the flow editor.

Example: 

image

The content element ids are saved in the node definition (same logic as transitions)

image

It can be specified in the flow generator function in the return object.

image

### DIFF
--- a/packages/studio-be/src/core/dialog/skill-service.ts
+++ b/packages/studio-be/src/core/dialog/skill-service.ts
@@ -21,7 +21,11 @@ export class SkillService {
     }
 
     // TODO change when studio is updated, since actual doesn't support catchall
-    return { flow: completeFlow, transitions: partialFlow.transitions }
+    return {
+      flow: completeFlow,
+      transitions: partialFlow.transitions,
+      ...(partialFlow.previewElements && { previewElements: partialFlow.previewElements })
+    }
   }
 
   private parseActionQuery(nodes): string[] | undefined {
--- a/packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
@@ -21,6 +21,15 @@ type Props = Pick<BlockProps, 'node'>
 const SkillCallContents: FC<Props> = ({ node }) => {
   return (
     <div className={cx(style.contentsWrapper, style['skill-call'])}>
+      {node.previewElements?.map((item, i) => {
+        return (
+          <ActionItem
+            key={`${i}.${item}`}
+            className={cx(style.contentWrapper, style.content, localStyle.item)}
+            text={`say #!${item}`}
+          />
+        )
+      })}
       {node.onReceive?.map((item, i) => {
         return (
           <ActionItem
--- a/packages/studio-be/src/sdk/botpress.d.ts
+++ b/packages/studio-be/src/sdk/botpress.d.ts
@@ -1184,6 +1184,8 @@ declare module 'botpress/sdk' {
     flow: SkillFlow
     /** An array of possible transitions for the parent node */
     transitions: NodeTransition[]
+    /** Elements to be previewed in the skill node */
+    previewElements?: string[]
   }
 
   /**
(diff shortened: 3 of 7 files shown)


davidvitora:
There is also the possibility of having a "preview" property instead. Here we can specify any kind of ActionItem to be previewed, including actions, which we can also preview today at normal nodes.

It would be something like 

`
preview: {
  type: sdk.NodeActionType.RenderElement,
  name: 'content-element-id-sdxsasf'
},
preview: {
  type: sdk.NodeActionType.RunAction,
  name: 'basic-skills/choice_parse_answer'
}
`

laurentlp (approved):
Tested and LGTM! @EFF can you please double-check!

packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
```diff
@@ -21,6 +21,15 @@ type Props = Pick<BlockProps, 'node'>
 const SkillCallContents: FC<Props> = ({ node }) => {
   return (
     <div className={cx(style.contentsWrapper, style['skill-call'])}>
+      {node.previewElements?.map((item, i) => {
+        return (
+          <ActionItem
```
EFF, line 26: is there any possibility that previewElements are not content element ?

packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
```diff
@@ -21,6 +21,15 @@ type Props = Pick<BlockProps, 'node'>
 const SkillCallContents: FC<Props> = ({ node }) => {
   return (
     <div className={cx(style.contentsWrapper, style['skill-call'])}>
+      {node.previewElements?.map((item, i) => {
```
EFF, line 24: what happens if previewElements is `undefined` ?  react will cry.
If the default value is null then please change the typing to

`previewElements: string[] | null` instead of `previewElements? string[]`

packages/studio-be/src/core/dialog/skill-service.ts
```diff
@@ -21,7 +21,11 @@ export class SkillService {
     }
 
     // TODO change when studio is updated, since actual doesn't support catchall
-    return { flow: completeFlow, transitions: partialFlow.transitions }
+    return {
+      flow: completeFlow,
+      transitions: partialFlow.transitions,
+      ...(partialFlow.previewElements && { previewElements: partialFlow.previewElements })
```
EFF, line 27: how is this serialized in the case if previewElements is null  or undefined ? 

`...undefined`

EFF (commented):
some questions need to be addressed before we move forward ?

packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
davidvitora: @EFF Not with the current implementation, but I left a comment above with another suggestion on how this additional use case can be implemented in the flow generator to preview an action, which I think can be useful for some use cases.

Let me know if you think about having that change to also allow actions

packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
davidvitora: I will double check that

packages/studio-be/src/core/dialog/skill-service.ts
davidvitora: With null or undefined 

````javascript
{
 flow: ...,
 transition:...
}

```` 

with previewElements defined

````javascript
{
 flow: ... ,
 transition: ... ,
 previewElements: ...
}

````

packages/studio-ui/src/web/views/FlowBuilder/diagram/nodes/SkillCallContents/index.tsx
davidvitora: Not a problem: node.onReceive? today can also be null or undefined

PR #390, opened 2022-08-22
feat(studio): allow third party content types to preview images

davidvitora:
image

Allow other content types which are not the "image" content type to render images in the preview

### DIFF
--- a/packages/studio-ui/src/web/util.ts
+++ b/packages/studio-ui/src/web/util.ts
@@ -63,3 +63,20 @@ export const sanitizeName = (text: string) =>
 export const stripDots = (str: string) => str.replace(/\./g, '--dot--')
 
 export const restoreDots = (str: string) => str.replace(/--dot--/g, '.')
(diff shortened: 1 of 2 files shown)


packages/studio-ui/src/web/utils.test.ts
```diff
@@ -0,0 +1,17 @@
...
+      micro: {
+        name: 'test',
+        type: 'node'
+      }
+    },
+    id: 2,
+    type: 'standard'
+  }
+  expect(recursiveSearch(obj, 'type')).toEqual(['image', 'node', 'standard'])
+})
```
laurentlp, line 17: Can you add the scenarios where `recursiveSearch` is called with an empty object, undefined, null, a number, ... Also, what if you are searching for a key that does not exist? Make sure to check all cases that you thought of handling when making the tests.

laurentlp (approved):
LGTM

PR #400, opened 2022-10-24
fix(ui): prevent long content type to overflow

EFF:
While fixing another issue, I had this very annoying overflow behaviour where very long content elements would overflow out of the drag n drop section and causing a hover flicker.

**Before**:
Screen Shot 2022-10-24 at 12 05 47 PM

Screen Shot 2022-10-24 at 12 31 43 PM

**After**:
Screen Shot 2022-10-24 at 12 03 54 PM
Screen Shot 2022-10-24 at 12 30 29 PM


davidvitora (approved):
Nice!!

PR #401, opened 2022-10-24
fix: properly set content type form step

EFF:
This PR fixes 2 issues that are mentioned [here](https://github.com/botpress/botpress/discussions/12141#discussioncomment-3939768)

1- It prevent to fetch invalid or undefined content element (the issue causing the 504)
2- Properly initializes the content type selector form when the category is already defined. (Issue is coming from the fact that I brought this piece of code from cloud studio which has a different behavior)

**before: forever spinning wheel**
Screen Shot 2022-10-24 at 12 13 47 PM

**before: create or select content item=**
Screen Shot 2022-10-24 at 12 12 06 PM


davidvitora (approved):
LGTM

PR #402, opened 2022-10-24
feat: double click on draggable item to edit

EFF:
Pretty simple and straight forward. This simple change allows a user to edit a node item by double clicking a draggable node item. One liner time saver for all bot builders


davidvitora (approved):
I tested and it's working fine

PR #405, opened 2022-11-04
feat(studio): multilingual captions for skill transitions

davidvitora:
Allows any skill to specify a caption for different languages; if not specified, use the default caption property

image

image

image

I also added the possibility of using the traditional intl module to translate. Example:

image

image

image

### DIFF
--- a/packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
+++ b/packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
@@ -5,6 +5,8 @@ import classnames from 'classnames'
 import _ from 'lodash'
 import Mustache from 'mustache'
 import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { RootReducer } from '~/reducers'
 import { restoreDots, stripDots } from '~/util'
 
 import style from './style.scss'
@@ -16,10 +18,11 @@ interface Props {
   position: number
   className?: string
   displayType?: boolean
-  getLanguage?: () => { currentLang: string; defaultLang: string }
+  currentLang: string
+  defaultLang: string
 }
 
-export default class TransitionItem extends Component<Props> {
+class TransitionItem extends Component<Props> {
   renderType = (): JSX.Element => {
     if (!this.props.displayType) {
       return null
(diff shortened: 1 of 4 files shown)


Sara-Reece (approved):

packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
```diff
@@ -126,3 +132,12 @@ export default class TransitionItem extends Component<Props> {
     }
   }
 }
+
+const mapStateToProps = (state: RootReducer) => ({
```
EFF, line 136: you can use the `withLanguage` HOC for this manner if you want but it's not necessary

packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
```diff
@@ -65,12 +69,14 @@ export default class TransitionItem extends Component<Props> {
...
+    const { position, transition, currentLang, defaultLang } = this.props
+    const { condition } = transition
 
-    if (caption) {
+    const caption = transition[`caption$${currentLang}`] || transition[`caption$${defaultLang}`] || transition.caption
+
+    if (caption && typeof caption === 'string') {
       const vars = {}
-      const htmlTpl = caption.replace(/\[(.+)]/gi, (x) => {
+      const htmlTpl = lang.tr(caption).replace(/\[(.+)]/gi, (x) => {
```
EFF, line 79: `lang.tr` looks up in the i18n translation files. are you sure you want to do this ?

EFF (commented):
I'm just not sure about this implementation.

packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
davidvitora: There are cases where we want to use translation files; for instance, for the onFailure string, it will not be dynamic, and we can use translation files instead to prevent node properties pollution

packages/studio-ui/src/web/views/FlowBuilder/common/TransitionItem.tsx
davidvitora: changed as suggested

EFF (approved):

